PR #105, opened 2026-07-07
feat(web): showcase community tweets alongside the demo video

peetzweg:
Extends the **See it in action** section from a lone YouTube embed into a full-bleed, horizontally snap-scrolling media strip: the demo video plus posts showing OpenDisplay in the wild ([@eduwass's triple-iPad setup](https://x.com/eduwass/status/2071902710597583300), [the Mactendo DS™ rig](https://x.com/peetzweg/status/2074416821738815692)).

**Why:** The demo video alone undersells it — real setups from the community are more spectacular, and more media will come over time.

**How:** A data-driven `Showcase` component — future media is one entry in its `ITEMS` array plus images in `public/showcase/`. Posts render as fully static, self-hosted cards (avatar, text, photo, date; the card links to the post on X). Deliberate call: static cards instead of live widgets.js embeds — the live embeds rendered unreliably under ad blockers and browser privacy features, and static cards also keep third-party tracking scripts off a page that advertises no telemetry. The strip breaks out of the content column to span the viewport; headings stay contained; snap-scroll with arrow controls instead of an auto-marquee so the video stays playable.

Verified with headless-Chromium screenshots of dev and built pages; the two pre-existing `tsc --noEmit` errors (TextRotate/main.tsx) are untouched.


peetzweg:
Follow-up: the strip is now full-bleed (headings stay in the column), and the tweets no longer look broken under ad blockers — widgets.js being blocked now falls back to styled quote cards with self-hosted copies of the tweet photos (`public/showcase/`). Verified both states headless: normal, and with `platform.twitter.com` + `*.twimg.com` DNS-blocked.

peetzweg:
Dropped widgets.js entirely — the posts are now fully static, self-hosted cards (avatar, text, photo, date, whole card links to X). The live embeds proved unreliable: browser privacy features tear the rendered embed down after insertion (the flash-then-fallback @peetzweg saw, reproducible in incognito), and ad blockers block the script outright. Static cards render identically for every visitor, sit in the prerendered HTML, and remove a third-party tracking script from a page that advertises no telemetry. Trade-off: no live engagement counts / Follow button.

PR #106, opened 2026-07-07
Add OpenDisplay Legacy: standalone iOS 12 companion app

lopsandrea:
## Summary

Adds a new, fully independent iOS Xcode target — **OpenDisplay Legacy** — for devices stuck on iOS 12 (iPhone 5s/6, iPad Air era) that can't run the existing SwiftUI/Combine-based companion app (both require iOS 13+).

- New target `OpenSidecariOSLegacy` (`deploymentTarget: "12.0"`), source folder `iOS-Legacy/`, no files shared with the modern `iOS/` target.
- UIKit throughout: classic `AppDelegate`, a `PhoneReceiverLegacy` port of the network/H.264-decode pipeline without Combine (delegate callbacks instead of `@Published`/`ObservableObject`), a UIKit `VideoView` for display + touch/gesture input, an idle/video-switching `RootViewController`, and a minimal `SettingsViewController` (device name, status, links).
- Same wire protocol as the existing Mac app (TCP :9000, length-prefixed H.264 Annex B frames, in-band JSON control messages) — **no Mac app changes required** for the core feature.
- MVP scope by design: no onboarding flow, no performance overlay, no Metal renderer toggle (all iOS-13+-only or experimental in the modern app).
- One small, separately-flagged Mac-side robustness fix included: `Mac/MacSender.swift`'s `SCShareableContent` lookup after creating a virtual display had a 5s timeout that measured too tight on at least one real machine — widened to 15s.

Several iOS-12-incompatible APIs were caught via real build/device verification and fixed along the way (see individual commits): `FileHandle.close()` → `closeFile()`, dynamic `UIColor.systemBackground`/`.systemGreen`/`.systemOrange`/`.secondaryLabel`/`.tertiaryLabel` → fixed light-mode equivalents, `NWTXTRecord` → hand-encoded `Data`-based TXT record, `UITableView.Style.insetGrouped` → `.grouped`, `UILaunchScreen: {}` → a real `LaunchScreen.storyboard`.

## Test plan

- [x] Builds successfully for `deploymentTarget: "12.0"` (`xcodebuild ... -destination 'generic/platform=iOS'`)
- [x] Installed and verified on a real iPad Air running iOS 12.5.8
- [x] Idle screen, Settings sheet (device rename, links), shake-to-open all verified on-device
- [x] Full streaming verified end-to-end against the unmodified Mac app: USB and WiFi connection, portrait/landscape rotation, device-name change reflected in the Mac's WiFi picker, background/foreground reconnect, touch-to-click and two-finger scroll
- [x] Manual code review pass (spec compliance + code quality) on every commit, plus a final whole-branch review comparing the new `PhoneReceiverLegacy` against the modern `PhoneReceiver` line-by-line for wire-protocol parity — no drift found

## Notes

During on-device testing over USB in one environment, intermittent connection drops surfaced (`CGVirtualDisplay creation failed` / `virtual display never appeared in SCShareableContent`). Investigation traced this to real OS-level USB detach/reattach events (confirmed via `remotepairingd`/`AMPDeviceDiscoveryAgent` system logs) rather than a code bug — WiFi was unaffected and works cleanly. Filed as a separate issue for visibility since it's a pre-existing Mac-app robustness question, not something this PR's scope covers fully.


peetzweg:
Interesting. Thanks for the contribution. Will be a bit of a bigger hassle to actually bring it to the AppStore but could revive a lot abandoned iPads.

peetzweg (commented):
Thanks for this — genuinely impressive work, especially the whole-branch line-by-line parity pass against `PhoneReceiver`. I compared `PhoneReceiverLegacy.swift` and `VideoView.swift` against the modern equivalents directly and the port is careful: listener lifecycle, watchdog, length-prefixed deframing, Annex B/NALU splitting, clock-offset sync, and the full touch/scroll forwarding logic (including coalesced/predicted touches) are byte-for-byte equivalent to the modern app. Verdict: **approve**, with one real bug worth fixing before/soon after merge and a couple of things to flag for the maintainer to decide on, not to block this PR.

**Worth fixing**

- `iOS-Legacy/SettingsViewController.swift:59` — the "Connection" status row is computed once in `cellForRowAt` and never refreshes while the sheet is open, so if the Mac connects/disconnects while Settings is up the row goes stale. Copilot already flagged this in review; still present in the latest commit. Easiest fix: reload that row (or the whole status section) from the existing `didChangeConnected` delegate callback while the settings VC is presented.

- Wire protocol has no version/capability field anywhere (`hello` on either side, `Mac/MacSender.swift`, `PhoneReceiver(Legacy).swift`) — this is pre-existing, not introduced here, but adding a second independent receiver implementation doubles the number of places a future protocol change (new control message, new hello field) has to be updated in lockstep, with nothing to catch drift except manual review. Not asking for this PR to add versioning, just flagging that the "no drift found" claim in the description is only true because someone did a careful manual diff — worth a lightweight guard (even a shared fixture/constants file for message keys) before a third receiver shows up.

**Minor / maintenance cost note (not blocking)**

- Confirmed duplication is real but scoped sensibly: `PhoneReceiverLegacy.swift` is a full copy of `PhoneReceiver.swift` (minus Combine, PerfStats, Metal path), and `Log.swift` is copied verbatim except for two string literals. `Log.swift` in particular has zero iOS 13+ API surface — it could be a single shared source file referenced by both targets' `sources:` in `project.yml` instead of a duplicate, which would remove one file from the "keep two copies in sync" list for free. Not required now, just an easy win if you want to trim the duplication surface later.
- `Mac/MacSender.swift`'s `findSCDisplay` timeout bump (5s → 15s) is a small, independently justified, separately-flagged change — no objection, and it doesn't touch the arrangement-memory code from #116/PR#127 (verified via `git merge-base`; that only looked like a conflict when diffing against current `main` tip because this branch predates that merge).

**Build integration** — clean. `project.yml` diff is purely additive (new scheme + new target block), no existing `OpenSidecariOS` target settings touched, no CI workflow references a scheme/xcodebuild target so nothing to update there, and `.gitignore`/`README.md` changes are additive and self-contained.

No correctness issues found in the decode/render/lifecycle path — the iOS 12 API swaps (`FileHandle.closeFile()`, fixed UIColor constants, `.grouped` table style, `LaunchScreen.storyboard`, hand-encoded DNS-SD TXT record for the Bonjour service name) all look correct, and the TXT record hand-encoding is a valid single-entry length-prefixed DNS-SD record that `NWTXTRecord`'s subscript parsing on the Mac side will read the same as the iOS13+-native encoding.

peetzweg:
let me revisit this. Convinced a legacy app might be handy yes. Heard a lot of requests for this.

peetzweg:
Thank you for this, and sorry for how long it sat. The port is careful work and it changed my mind about serving the oldest iPads. In the end the toolchain decided the shape, not the code.

- Xcode 27's iOS SDK has a minimum deployment target of 15.0. App Store uploads must use the current year's Xcode from each April, so an iOS 12 app could ship until roughly spring 2027 and never receive an update after that.
- Given that, the main app now runs on iOS 15 as of 1.21.0 (https://github.com/peetzweg/opendisplay/pull/291), which covers everything from the iPad Air 2 and iPhone 6s up. iOS 13 and 14 add no devices over 15.
- The devices this PR was for (iPad Air 1, mini 2 and 3, iPhone 5s and 6) stay on iOS 12 and can only be served by a sideloaded build. The clean-room client in https://github.com/cuongpham1/ipad-second-monitor-ios12-free speaks the current protocol and is the best home for that; it was written against PROTOCOL.md without touching this repo.
- The SCShareableContent timeout widening this PR also carried was superseded by the bounded recovery in #288.

Closing without merging. If you want to point iOS 12 users somewhere from the README's compatible apps list, a PR adding a line there is very welcome.

PR #118, opened 2026-07-08
fix(mac): log rejected encodes instead of dropping them silently

lotgood:
## What

The encode output handler used to bail on any failure without a trace:

```swift
guard status == noErr, let buffer, let self else { return }
```

This adds a rate-limited failure counter/log (first failure, then every 120th) so a wedged encoder shows up in `/tmp/*.log` within a second.

## Why

A session that rejects every frame is indistinguishable from a healthy-but-idle pipeline: capture counters keep counting, no error surfaces anywhere, the receiver just stays black.

This is not hypothetical. H.264 has a hard level-5.2 pixel-rate ceiling (~535M luma samples/s). Feed the encoder frames above it and `VTCompressionSessionEncodeFrame` returns **noErr** while handing the output handler a **nil buffer for every single frame** — measured on a private high-refresh experiment: a large capture at 120fps produced zero output with zero errors, and diagnosing it took hours precisely because this guard ate the evidence. Upstream doesn't drive >60fps today, but any future resolution/rate bump (or an encoder property mismatch, or resource exhaustion) hits the same silent black screen.

## Testing

- On the build where this was found, the log line was what turned "receiver stays black, everything looks healthy" into "encode output failed: status=0 buffer=false" — immediate diagnosis.

Sibling PRs: #117 and #119 — trivial conflicts if merged together, happy to rebase.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -222,6 +222,9 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // Encode failures repeat every frame once the session goes bad; throttle
     // the log to one line a second and carry the count.
     private var encodeFailureLogPolicy = ThrottledLogPolicy<OSStatus>()
+    // Same for the encoder output callback rejecting a frame; separate policy
+    // so "submit failed" and "output rejected" stay distinguishable.
+    private var encodeOutputFailureLogPolicy = ThrottledLogPolicy<OSStatus>()
     // A framing desync feeds this garbage at the peer's message rate until the
     // watchdog redials, so it needs the same treatment. Detail is the byte
     // count of the last message that would not parse.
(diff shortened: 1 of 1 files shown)


peetzweg (commented):
Thanks for this — good catch, and a genuinely nasty failure mode (`noErr` + nil buffer on the level-5.2 ceiling is exactly the kind of thing that looks like a healthy idle pipeline). This is a small, well-scoped diff and it does what the title says. Verdict: approve, with one thing worth fixing before/at merge.

**Worth fixing** — `Mac/MacSender.swift` (the new `if status != noErr || buffer == nil` block, `encodeFailures += 1` / the modulo check): this mutates a plain `Int` from inside the `VTCompressionSessionEncodeFrame` output handler. Per #119's own description, that handler runs on VideoToolbox's thread, not on `self.queue` — and #119 explicitly hops back to `self.queue` for its `pendingEncodes` decrement for exactly this reason. If multiple encodes are ever in flight (the scenario #119 is built to handle), concurrent callbacks can race on `encodeFailures`, so the modulo gate isn't reliable under the exact saturation conditions this PR is meant to catch. Low severity — it only skews a debug counter, doesn't touch stream correctness — but worth the same one-line queue hop for consistency with the sibling fix.

**Minor** — `status=\(status)` logs the raw `OSStatus` int rather than a named error. Not a blocker: I checked, there's no OSStatus-to-name helper anywhere else in the codebase (the `VTCompressionSessionCreate failed` log doesn't print a status either), so this matches existing convention. A small decoder for common VT codes would be a nice follow-up, not a requirement here.

On the thing I was most worried about — log flooding on a saturated encoder — this is actually handled well: incrementing on every failure is cheap, and gating the `Log.info` call to every 120th occurrence means a fully saturated 60fps encoder logs roughly once every 2 seconds, not 60 lines/sec. No hot-path cost from this change.

Noting the overlap with #117 and #119 (same file, adjacent lines) per the author's heads-up — nothing here needs to block on those, just expect a rebase.

peetzweg:
Thanks for this one too — the failure mode is real, and the diagnosis story ("receiver stays black, every other counter looks healthy") convinced me. The nil-buffer-despite-noErr case in particular is worth surfacing: nothing else in the pipeline can see it.

The file moved a lot since July, so rather than ask you to rebase across it I pushed to your branch directly (maintainer edits):

- rebased onto current main
- kept your guard-and-log approach in the encode output handler, but routed it through the `ThrottledLogPolicy` infrastructure main gained in the meantime, instead of the hand-rolled `% 120` counter — the synchronous submit path (`submitStatus != noErr`) already reports through a policy, and the output handler now mirrors it one-to-one (separate policy instance, so "submit failed" and "output rejected" stay distinguishable in the log)
- the report line calls out the nil-buffer case explicitly: `encoder output rejected: nil buffer despite noErr (N since last report)`

Build is clean and the existing `ThrottledLogPolicy` tests cover the throttling behaviour. Shout if you'd rather it worked differently — happy to adjust.

PR #119, opened 2026-07-08
fix(mac): cap in-flight encodes so a saturated encoder sheds frames, not latency

lotgood:
## What

Adds a `pendingEncodes` gate to `MacSender`: frames handed to `VTCompressionSessionEncodeFrame` whose output handler hasn't fired yet. While 2 are already in flight, capture frames are skipped (counted as drops, like the existing socket backpressure path).

## Why

`VTCompressionSessionEncodeFrame` accepts frames faster than the hardware encoder can process them and simply queues the excess inside VideoToolbox. Nothing in the pipeline notices — `pendingSends` only guards the socket — so when capture outpaces the encoder, every queued frame is pure glass-to-glass latency.

Measured on a private 120Hz experiment on top of this pipeline: capture→socket 62ms under sustained load vs 14ms idle — exactly ~5 queued frames at ~13ms per encode. With the gate, delivered fps becomes whatever the encoder actually sustains, at ~1 frame of queueing latency instead of 5. At upstream's 60fps the window is smaller but the failure mode is the same whenever an encode takes longer than a frame interval (large captures, thermal throttling, encoder contention with other apps).

Skipped frames are pre-encode drops, so the P-frame chain is unaffected.

Robustness details:
- gate resets on encoder rebuild and reconnect, so callbacks swallowed by an invalidated session can't wedge it shut
- the synchronous-failure path decrements with `max(0, ·)` for the same reason
- the handler decrement hops to the sender queue (the handler runs on VideoToolbox's thread)

## Testing

- Verified on the experimental build where this was found: under encoder saturation, latency holds at ~1 frame while fps settles at the encoder's sustainable rate.

Sibling PRs: #117, #118 — trivial conflicts if merged together, happy to rebase.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -117,6 +117,9 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // Kept tight: at 60fps each queued send is ~17ms of added latency.
     private var pendingSends = 0
     private let maxPendingSends = 3
+    // Encoder saturation gate (see the capture callback): frames handed to
+    // VTCompressionSession whose output handler hasn't fired yet.
+    private var pendingEncodes = 0
     private var dropsThisWindow = 0
     private var needsKeyframe = true
     private var connectionReady = false
@@ -568,6 +571,7 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         connection?.cancel()
         connection = nil
         pendingSends = 0
+        pendingEncodes = 0
         queue.asyncAfter(deadline: .now() + 1.0) { [weak self] in
             self?.connect()
         }
@@ -801,6 +805,9 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // MARK: - Encoder setup
 
     private func setupEncoder(width: Int, height: Int) {
+        // Reset the in-flight gate on every (re)build so callbacks the old
+        // session swallowed can't wedge it shut.
+        queue.async { self.pendingEncodes = 0 }
         // Low-latency rate control: the hardware encoder emits every frame
         // immediately instead of pipelining. (`-lowlatency NO` for A/B.)
         let lowLatency = UserDefaults.standard.object(forKey: "lowlatency") == nil
@@ -862,6 +869,19 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
             return
         }
 
+        // Encoder saturation gate: a hardware encode is not free, and
+        // VTCompressionSessionEncodeFrame happily QUEUES what the encoder
+        // can't keep up with — queued frames are pure latency (measured on a
+        // 120Hz fork of this pipeline: capture→socket 62ms under sustained
+        // load vs 14ms idle, exactly ~5 queued frames). Keep at most 2 in
+        // flight and skip the rest: delivered fps becomes whatever the
+        // encoder sustains, at ~1 frame of latency instead of 5.
+        if pendingEncodes >= 2 {
+            dropsThisWindow += 1
+            dropsTotal += 1
+            return
+        }
+
         encode(pixelBuffer, pts: CMSampleBufferGetPresentationTimeStamp(sampleBuffer))
     }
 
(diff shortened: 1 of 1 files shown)


peetzweg (commented):
Thanks for this — the write-up and the measured numbers (62ms vs 14ms, ~5 queued frames) make the motivation easy to follow, and the thread-hopping in the completion handler is handled correctly. Overall this is solid and I'm comfortable merging it; I have a couple of non-blocking robustness points worth a look, plus a real conflict with #118 to sort out before both land.

**Verdict: approve, with a couple of things worth addressing (not blocking).**

Worth fixing:

- `Mac/MacSender.swift:807-810` — the reset in `setupEncoder` (`queue.async { self.pendingEncodes = 0 }`) races against stray completions from the *old*, just-invalidated `VTCompressionSession`. Those old callbacks decrement on `queue` too, and there's no ordering guarantee that all of them land before the reset's `queue.async` runs. If a stray old-session decrement arrives after the reset (once the new session already has 1-2 real encodes in flight), it undercounts the new session and lets one extra frame through the gate — self-healing, not a wedge, but it does mean "at most 2 in flight" isn't quite a hard invariant across a reconfigure. This file already has a pattern for exactly this class of bug — `dialGeneration` guards stale USB-dial callbacks from a superseded attempt. Worth doing the same here (e.g. an `encoderGeneration` bumped in `setupEncoder`/`stop`, checked before applying a decrement) rather than relying on `max(0, ·)` to paper over it.
- `Mac/MacSender.swift:920-924` — the comment ("keeps a rare double-decrement from wedging the gate shut") has the failure mode backwards: a double-decrement *undercounts* pendingEncodes, which makes the gate more permissive, not shut. The real "wedge it shut forever" risk is a **leaked increment** (encode started, neither the sync-failure path nor the completion ever fires), and that's actually handled by the explicit resets on rebuild/reconnect, not by this clamp. Worth fixing the comment so the next person reading it doesn't reason about the wrong failure mode.
- `Mac/MacSender.swift:879` — cap is a bare `2`, unlike `maxPendingSends` a few lines up which is a named `let`. Minor, but pull it into a constant for consistency and so the "at most 2 in flight" claim in the description is grep-able from one place.

Minor:

- `Mac/MacSender.swift:872-883` — the encoder-saturation drop shares `dropsThisWindow`/`dropsTotal` with the socket-backpressure drop just above, so the phone-side stats can't tell "socket backed up" from "encoder saturated" apart. Not this PR's job necessarily (and #118 is already adding encode-failure telemetry nearby), but flagging since you'll want that distinction the first time someone asks "why is fps low."
- Confirmed the keyframe-forcing frame isn't exempted from the new gate: if `needsKeyframe` is set (reconnect, phone's `"kf"` request) while `pendingEncodes >= 2`, that frame gets dropped like any other and the flag stays sticky for the next one that gets through. No actual keyframe loss since `needsKeyframe` isn't cleared on this drop path, just a small delay proportional to saturation — fine given the gate drains every ~13ms, just wanted to call out that I checked it since it's exactly the scenario I was worried about.
- Verified against the real failure mode #118 documents (VT returning `noErr` with a nil buffer at the H.264 level-5.2 pixel-rate ceiling): this PR's completion handler decrements `pendingEncodes` unconditionally *before* the `status == noErr` check, so that path does not leak a permit. Good — that was my top suspicion going in and it holds up.

Conflict to flag: #118 rewrites the same completion-handler guard this PR touches (`guard status == noErr, let buffer, let self else { return }` → the multi-statement failure-logging version), so merging both will produce a real, not just proximity, conflict — whoever rebases needs to make sure the unconditional `pendingEncodes` decrement still runs first, ahead of #118's failure-logging branch. #117 only touches the `pendingSends` block just above this PR's insertion point, so that one's a trivial context conflict as you said.

Nice fix — the pre-encode drop reasoning (P-frame chain untouched since the frame never reaches the encoder) is correct, and the gate itself is properly serialized on `queue` for every increment/decrement path I traced (capture callback, watchdog keyframe replay, completion handler's `queue.async` hop, and both explicit resets).

jlfwong:
FWIW, in optimizing a fork of this for iPad drawing support, I also ran into this same performance gotcha.

I was hitting encode times of >100ms/frame. They were pipelined & parallelized, so I was still getting 40fps, but in the case of drawing apps, very low latency is important independent of fps. After preventing parallel encodes, it decreased end-to-end latency by 60ms/frame for me when running at native iPad HiDPI resolution. The result was viscerally obvious.

It isn't clear to me that there's any benefit of having multiple encodes in flight at once, so I set mine to drop frames whenever there's an encode pending (effectively `maxPendingEncodes = 1`)

peetzweg:
Closed in favor of https://github.com/peetzweg/opendisplay/pull/162

Thanks for spotting this earlier @lotgood ! 🙏 
I just came around now to do some reviews and proper testing again.

Issue #116, opened 2026-07-08
Remember screen arrangement after reconnecting


christophby:
After reconnecting the same iPad, macOS doesn't remember the window arrangement I set up before.

Great app, btw <3

peetzweg:
Thanks for the report — confirmed and fixed in #127.

What was happening: macOS does remember display arrangement, but keys it on the monitor's identity — and OpenDisplay's virtual display identity changed whenever the iPad rotated or you switched between USB and WiFi. macOS then treated it as a brand-new monitor and parked it at the default spot right of your main display.

The app now remembers the position itself, keyed on the device, and puts the display back where you left it — across reconnects, rotations, and transport switches. The latest spot you drag it to wins, whichever orientation you're in.

It'll be in the next released build, and earlier on TestFlight if you'd like to try it sooner: https://testflight.apple.com/join/3NYaY11c

PR #121, opened 2026-07-09
Draft: explore Android receiver and Chinese UI fork work

chenyuhao0628:
## Summary

This is a draft PR for discussion, not a request for immediate merge.

It packages the Android/CN fork work into a visible upstream thread so the
project can evaluate whether any parts should be split into smaller upstreamable
PRs.

## What changed

- Adds an Android WiFi receiver prototype that follows the existing OpenDisplay
  receiver contract:
  - `_opensidecar._tcp` service advertisement
  - TCP listener on port 9000
  - length-prefixed JSON control messages
  - H.264 Annex B frame reception
  - MediaCodec rendering to SurfaceView
  - cursor overlay, tap/drag input, and two-finger scroll
- Adds Chinese user-facing copy for Mac and iOS UI surfaces.
- Makes the Mac app easier to run as a visible Dock/window app.
- Fixes mirror-mode input setup so touch input is not limited to extend mode.
- Adds fork documentation for architecture, roadmap, contribution, support, and security.

## Why draft

The branch intentionally contains several related pieces from a working fork.
For upstream review, this probably needs to be split into smaller PRs:

- Android receiver
- Mac sender protocol/input changes
- localization changes
- documentation changes

Opening this as a draft keeps the work discoverable without implying that the
full branch should be merged as-is.

## Validation

Verified locally:

- macOS Debug build for `OpenSidecarMac`
- iOS Simulator Debug build for `OpenSidecariOS`
- Android debug APK build
- Android protocol self-test
- real Android tablet WiFi mirror/extend session with cursor display and touch stability

## Notes

This work is based on the fork at:

https://github.com/chenyuhao0628/opendisplay-mac-android-cn


peetzweg (commented):
Thanks for opening this as a draft and for being upfront that it's a fork dump rather than a merge request — that makes this easier to review honestly. There's clearly real, working effort here (the Android receiver runs against the existing protocol, and a few of the Mac-side fixes are legit bugs). Here's my early, directional feedback, organized around the four things that matter before any of this becomes a mergeable PR.

## a) Duplication with #130

`AndroidReceiver/` here overlaps almost entirely with #130, and #130 already accounts for it: it explicitly imports the Android receiver "from #121 by @chenyuhao0628", keeps your authorship on the introducing commit, and notes the fork's GPL-3.0 licensing. So the coordination problem is already half-solved — what's missing is you two actually talking. Concretely, #130 also fixes the two biggest problems I'd otherwise flag in this PR's Android code:

- `AndroidReceiver/app/src/main/java/app/opendisplay/android/MainActivity.java` has all UI copy as hardcoded Chinese string literals (`"等待启动…"`, `"设置与帮助"`, etc.) with no `res/values/strings.xml` at all. #130 extracts these into resources with English as the default locale and a `values-zh` overlay.
- #130 also adds a reconnect/keepalive fix for stale TCP connections on the Android side that this PR doesn't have.

My suggestion: don't try to land `AndroidReceiver/` from this branch. Go review/co-author #130 instead — your touch/scroll/cursor-overlay work is already credited there, and duplicating a second Android receiver in-tree serves nobody. If there's something in this version's `TouchGestureCoordinator.java` / `CursorOverlayView.java` that #130 doesn't have, call it out as a comment on #130 rather than carrying two implementations forward here.

## b) Localization: this is a hard fork, not a mergeable l10n change

The Chinese UI work replaces English string literals in place, directly in Swift source — `Mac/MacSender.swift`, `Mac/OpenSidecarMacApp.swift`, `iOS/OpenSidecarPhoneApp.swift`, `iOS/PhoneReceiver.swift` all have their English `Text("…")` / `status(…)` literals rewritten to Chinese. There's no `Localizable.strings`, String Catalog (`.xcstrings`), or `NSLocalizedString`/`String(localized:)` involved — merging this as-is would make the shipped app Chinese-only for every user, not an added locale.

For this to be mergeable, it needs to go through Xcode's String Catalog mechanism: wrap the user-facing strings, generate a catalog with English as the base language, add a `zh-Hans` translation layer on top, and leave the default behavior (English) untouched. #130's `values` + `values-zh` split on the Android side is exactly the right shape for this — worth mirroring it here.

## c) Risky/broken changes to shared Mac/iOS code

A few things here would break the app or regress recently-shipped behavior if merged as-is:

- **`Mac/DisplayArrangement.swift` is deleted outright**, and `Mac/VirtualDisplay.swift` drops `restoreOrigin`/`onOriginChange`/`manageOrigin` entirely, along with the arrangement-key wiring in `MacSender.swift`. That's the display-arrangement-persistence feature from #116 (PR #127), merged into `main` after this branch was forked. This isn't an intentional change — it's the branch predating that merge. It needs a rebase onto current `main`, not a manual reconciliation, or arrangement memory silently regresses for every user.
- **`project.yml` changes the iOS bundle identifier** from `com.peetzweg.opensidecar.ios` to `com.cyh.opendisplay.receiver`. That's almost certainly a personal-signing artifact from building the fork locally, but if it landed it would break the app's identity/signing/App Store continuity. This has to be reverted before anything here is mergeable.
- **`LSUIElement: true` is removed from the Mac Info.plist**, and `SenderController`'s default `presentation` flips from `.menuBar` to `.dock` (`Mac/OpenSidecarMacApp.swift`). That silently changes default behavior for every existing user (Dock icon always visible) and is bundled in with unrelated l10n/doc changes. If this is wanted, it should be its own explicit, discussed change with the reasoning spelled out, not smuggled in alongside string replacements.
- **`script/LauncherMain.swift` hardcodes `/Users/cyh/Documents/opendisplay`** as the project root. Fine for your personal fork, not something that can go into shared tooling as-is.

On the positive side, two things look like genuinely good, small, independently-mergeable fixes and I'd like to see them split out on their own:
- The mirror-mode touch-input fix in `MacSender.swift` (`inputInjector` now gets wired up in the `.mirror` case too, plus the `waitForAccessibilityIfNeeded` refactor) — this is a real bug fix (touch was extend-only before) and reads cleanly.
- `Mac/Usbmux.swift`'s new `OneShotResume` — replacing an unsynchronized `resumed` bool with a lock-guarded resume gate around the continuation is a legitimate thread-safety fix for a real race.
- One thing to double check when that fix moves to its own PR: `MacSender` now conforms to `@unchecked Sendable` — worth confirming that's actually justified (what data is shared across isolation domains) rather than just silencing a Swift 6 concurrency warning.

Also worth a mention: `README.md`, `ARCHITECTURE.md`, `ROADMAP.md`, `CONTRIBUTING.md`, `SECURITY.md`, `SUPPORT.md` are all rewritten wholesale as fork-branded docs ("OpenDisplay Mac/Android CN", pointing at your fork repo). Those should stay in your fork, not overwrite upstream's project docs.

## d) What I'd want as separate, reviewable PRs

1. **Android receiver** — redirect this energy to #130 rather than a competing PR; coordinate there.
2. **Mirror-mode touch-input fix + Usbmux resume-race fix** — small, standalone, rebased on current `main`, no l10n or doc changes mixed in. This one I'd probably take almost as-is.
3. **Localization**, done via String Catalog / resource strings with English as default and Chinese as an added locale — not a full source-level string replacement.
4. Drop from anything upstream-facing: the bundle ID change, the `LSUIElement`/Dock-default flip (unless proposed and discussed on its own), the personal-path launcher script, and the fork-branded doc rewrites.

Appreciate you putting this up as a draft rather than a surprise merge request — once it's rebased onto current `main` and split along these lines, the Mac-side fixes in particular should be quick to land. Happy to look again once any of these are peeled off into their own PRs.

Issue #111, opened 2026-07-07
Stop macOS classifying the virtual display as a TV (root cause of #100 mirror default)


peetzweg:
Follow-up to #100 (root cause). #100 shipped a robust guard — the extend-mode virtual display is forced out of any system mirror set — so the user-facing symptom (stuck mirroring) is handled regardless of classification. This issue is the *cleaner* fix: stop macOS misclassifying the virtual display as a **TV** in the first place, which is what triggers the "Mirror Entire Screen" default (and the 1×-instead-of-@2x resolution seen in the reporter's screenshot).

## Background

macOS synthesizes an **EDID** for the `CGVirtualDisplay` and classifies it as monitor vs. TV from that (TVs carry a CEA-861 extension; the "When connected to TV" row + TV icon are macOS reflecting that guess). We can't write the EDID or set a "this is a monitor" flag — the private descriptor (`Mac/CGVirtualDisplayPrivate.h`) only exposes `name`, `maxPixelsWide/High`, `sizeInMillimeters`, `serialNum`, `productID`, `vendorID`, and the mode list. So the only levers are those inputs.

**The hard constraint:** the misclassification is **macOS-version-specific**. On the maintainer's Mac (Apple M5 Pro, macOS 26.5.1) the *exact same descriptor* is classified as a **monitor** and extends perfectly; only the reporter's setup (iPad Air 4, iPadOS 26.5.2, macOS version TBD) misclassifies it. So any descriptor change here **cannot be validated on the maintainer's machine** — it needs the reporter, or a macOS version that reproduces. This is why it's `agent:explore`, not `agent:ready`.

## Concrete lead: compare to DeskPad (a CGVirtualDisplay that is NOT classified as a TV)

DeskPad ([Stengo/DeskPad](https://github.com/Stengo/DeskPad), `DeskPad/Frontend/Screen/ScreenViewController.swift`) uses the same private API and shows up as a normal monitor. Its descriptor vs. ours (`Mac/VirtualDisplay.swift:25-44`):

| Field | OpenDisplay | DeskPad |
|---|---|---|
| `sizeInMillimeters` | **147×68** (phone-sized; wrong for an iPad) | 1600×1000 (large, generic) |
| `productID` / `vendorID` | 0x4F53 / 0x5043 (ASCII "OS"/"PC") | 0x1234 / 0x3456 (generic) |
| `maxPixelsWide/High` | sized to the device panel | 5120×2160 |
[...]

Issue #100, opened 2026-07-05
Stuck in mirror mode: virtual display detected as a TV → macOS forces "Mirror Entire Screen"


peetzweg:
## Summary

A user reports the device only ever shows a **mirror** of the Mac's main screen, no matter what they change. The OpenDisplay app is correctly set to **Extend** and reports "Extending to iPad" — but macOS itself is mirroring the virtual display. The screenshots show the root cause: **macOS is classifying the CGVirtualDisplay as a _television_, and TV-class displays default to "When connected to TV → Mirror Entire Screen"**, which puts the virtual display into a mirror set with the built‑in display. The app's Extend/Mirror toggle has no effect on this because the mirroring is imposed by macOS's display arrangement, and the app never actively forces the display out of a mirror set.

Reported on Reddit: https://www.reddit.com/r/ProductivityApps/comments/1un1342/comment/ovgof1p/

## Environment

- **Device:** iPad Air (4th generation), iPadOS 26.5.2
- **Transport:** WiFi
- **App mode:** Extend (confirmed in the app UI), Quality: Best (native), all permissions granted (Screen Recording, Accessibility, Local Network)
- **Mac / macOS version:** not stated by the reporter — the Settings UI is macOS 15 Sequoia or later. *(Please confirm — see open questions.)*

## What the reporter sees

- The iPad shows a mirror of the Mac's main display regardless of settings.
- In **System Settings → Displays**, the OpenDisplay device:
  - is drawn with a **TV icon** (not a monitor/iPad icon),
  - has **"Use as" = Main display**,
  - shows a **"When connected to TV" = "Mirror Entire Screen"** row (this row only appears for displays macOS thinks are TVs),
  - lists **1×/native-pixel** resolutions ("2360 × 1640 (Default)", 1600×1112, 1180×820) rather than a clean @2x HiDPI default.
- **"When screen sharing is started, the resolution changes on the Mac."**

*(Reporter screenshots of both the OpenDisplay app and System Settings → Displays are in the Reddit thread and can be attached here.)*

## Diagnosis (high confidence)

[...]

PR #126, opened 2026-07-10
fix(mac): virtual display hidden from ScreenCaptureKit by the TV mirror default

peetzweg:
Second report on #100; also touches #111. **Parked as draft: the two commits have very different risk profiles and we haven't decided yet how much of this to ship — see "Ship decision" below.**

**Why:** After 0.11.0 the reporter's failure changed from "stuck mirroring" to `WiFi • Failed: virtual display never appeared in SCShareableContent`. Reproduced the mechanism locally with a harness that force-mirrors the virtual display at creation (what the TV default does on the reporter's Mac):

1. A **mirror-set secondary is not an active display and is omitted from `SCShareableContent` entirely** — the display doesn't show up wrong, it vanishes from ScreenCaptureKit's world.
2. macOS restores the saved TV "Mirror Entire Screen" arrangement **asynchronously, seconds after creation**. The #109 guard detaches it on its next tick (≤2s cadence), SCK takes ~2s more to re-list the display.
3. That round trip regularly overruns `findSCDisplay`'s hard 5s window → the exact error in the report.

## Commit 1 — robustness + diagnostics (safe, inert for healthy machines)

- `findSCDisplay` waits 15s instead of 5s (same direction PR #106 proposed). Healthy machines still exit on the first successful poll (~0.3–3s measured) — nobody waits longer unless they're already failing.
- While waiting, log the CG-level display state, and embed it in the timeout error (`inMirrorSet/mirroring/active/online/main`). If this still fails for the reporter, **the next screenshot diagnoses itself** — worth more than knowing their macOS version.
- Bug fix in `ensureNotMirrored`: it scanned the **active** display list to find displays mirroring the VD, but a mirroring display is never active (measured) — switched to the **online** list. This code path only executes when a VD is already in a mirror set, i.e. never on unaffected Macs.

Based on the measured timings there's a decent chance this half alone fixes the reporter.

## Commit 2 — root-cause experiment (#111): monitor-like identity (speculative, affects everyone)

Mirrors what DeskPad (same private API, never TV-classified) does differently: a ladder of standard modes (1024×768/800×600/640×480 when they fit under native), PPI-derived physical size instead of the phone-sized 147×68mm constant, and a product-id bump (`"OS"`→`"OD"`) so installs with a poisoned saved mirror arrangement start from a clean slate.

Costs, honestly: **every existing user's Mac forgets the display's saved position once** (identity change); System Settings lists three extra selectable resolutions the enforcement loop will fight; and the classification benefit is **unvalidated** — the maintainer's machine classifies the display correctly, so this cannot be tested without the reporter's setup.

## Ship decision (open)

Options, in rough order of preference:
1. **Merge commit 1 only**; park commit 2 on this branch, linked from #111, until the reporter's data (macOS version + the new self-diagnosing error text from a commit-1 build) proves classification is still the blocker.
2. Commit 1 + commit 2 gated behind a hidden `defaults` knob; hand the reporter a Terminal one-liner to opt in, promote once proven.
3. Ship both as-is.

## Verification

Fake receiver + mirror-attacker harness: session heals to "Extending" in ~1.3s under attack; both rotation rebuilds clean; baseline extend unaffected. Verified on macOS 26.5.1 (M5 Pro) — which does *not* reproduce the TV classification, hence the caution above.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -251,10 +251,15 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         // = native pixels / 2 (rounded down to even for the encoder).
         let pointsWide = (info.pixelsWide / 2) & ~1
         let pointsHigh = (info.pixelsHigh / 2) & ~1
-        // Rough physical size so macOS picks a sane default UI scale.
-        let mm = info.pixelsWide >= info.pixelsHigh
-            ? CGSize(width: 147, height: 68)
-            : CGSize(width: 68, height: 147)
+        // Physical size from a per-kind PPI estimate (iPads ≈264, iPhones
+        // ≈460) so macOS picks a sane default UI scale. The old 147×68mm
+        // constant described a phone even for iPads — a wildly wrong size
+        // feeds the display-class heuristic bad data (#111), and TV-classed
+        // displays default to "Mirror Entire Screen" (#100). Orientation is
+        // inherent: pixelsWide/High arrive swapped on rotation.
+        let ppi = info.kind == "iPad" ? 264.0 : 460.0
+        let mm = CGSize(width: Double(info.pixelsWide) / ppi * 25.4,
+                        height: Double(info.pixelsHigh) / ppi * 25.4)
 
         // USB sessions can start before lockdown resolves the device name —
         // fall back to the kind from the hello rather than the generic label.
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -329,17 +329,42 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         }
     }
 
-    /// The virtual display takes a moment to show up in shareable content.
+    /// The virtual display takes a moment to show up in shareable content —
+    /// and on machines that misclassify it as a TV (#100), macOS restores the
+    /// saved "Mirror Entire Screen" arrangement asynchronously, seconds after
+    /// creation. A mirror-set secondary is not an active display and is
+    /// omitted from SCShareableContent entirely, so the display *vanishes*
+    /// until the un-mirror guard (`VirtualDisplay.ensureNotMirrored`, ≤2s
+    /// cadence) detaches it, plus ~2s more to re-register with SCK. That
+    /// round trip regularly overran the old 5s window (also reported
+    /// independently in PR #106) — allow 15s, and log the CG-level state
+    /// while waiting so a failure pinpoints why SCK isn't listing it.
     private func findSCDisplay(id: CGDirectDisplayID) async throws -> SCDisplay {
-        for _ in 0..<20 {
+        for attempt in 0..<60 {
             let content = try await SCShareableContent.current
             if let display = content.displays.first(where: { $0.displayID == id }) {
                 return display
             }
+            if attempt % 4 == 3 {
+                Log.info("display \(id) not in SCShareableContent yet: \(Self.displayState(id))")
+            }
             try await Task.sleep(for: .milliseconds(250))
         }
         throw NSError(domain: "MacSender", code: 3,
-                      userInfo: [NSLocalizedDescriptionKey: "virtual display never appeared in SCShareableContent"])
+                      userInfo: [NSLocalizedDescriptionKey:
+                          "virtual display never appeared in SCShareableContent (\(Self.displayState(id)))"])
+    }
+
+    /// One-line CG-level view of a display for diagnosing why SCK omits it
+    /// (mirror-set members that aren't the set's primary are inactive and
+    /// hidden from SCShareableContent — the #100 failure mode).
+    private static func displayState(_ id: CGDirectDisplayID) -> String {
+        let mirrors = CGDisplayMirrorsDisplay(id)
+        return "inMirrorSet=\(CGDisplayIsInMirrorSet(id) != 0)"
+            + (mirrors == kCGNullDirectDisplay ? "" : " mirroring=\(mirrors)")
+            + " active=\(CGDisplayIsActive(id) != 0)"
+            + " online=\(CGDisplayIsOnline(id) != 0)"
+            + " main=\(CGDisplayIsMain(id) != 0)"
     }
 
     private func startCapture(display: SCDisplay, pixelsWide: Int, pixelsHigh: Int) async throws {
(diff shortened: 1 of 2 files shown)


peetzweg (commented):
Commit 1 (`ea3fa34`, #100) does what it claims and should ship; commit 2 (`3a7ea30`, #111) is exactly as speculative as the PR body admits and I'd hold it. Overall verdict: **approve commit 1, do not merge commit 2 yet** — go with option 1 from the "Ship decision" section.

## Tracing the fix against the failure mode

The reporter's arrangement was "VD set as Main, built-in mirroring it." Before this PR, `ensureNotMirrored` broke that by (a) detaching the VD's own mirror-of setting and (b) scanning `CGGetActiveDisplayList` for any *other* display mirroring the VD and detaching that too. But a display that is mirroring another is not active — so step (b) could never find the built-in in exactly this arrangement, silently no-op'ing on the reporter's specific setup since #109 shipped. `Mac/VirtualDisplay.swift:148-150` switches that scan to `CGGetOnlineDisplayList`, which does include it. This is the load-bearing line of the PR: without it, the un-mirror guard only ever fixed "VD mirroring the built-in," not "built-in mirroring the VD," and the second is what the reporter has. Good catch, and it's correctly scoped — matching only `CGDisplayMirrorsDisplay(other) == id` means it can't touch mirror sets that don't involve the VD.

The `findSCDisplay` timeout bump (`Mac/MacSender.swift:347-361`, 5s → 15s) is arithmetically justified by the guard's own cadence (≤2s to detach + ~2s for SCK to re-list), and is inert for healthy machines since they exit the loop on first success. Combined with the online-list fix, this covers the reported failure mode.

## Worth fixing

- **`findSCDisplay`'s retry loop doesn't check `stopped`** (`Mac/MacSender.swift:348-361`). Every other retry/poll loop in this file guards on `stopped` (`start()` at 206/237, `reconfigure` at 309/313, `connect` at 473, the reconnect scheduler at 541/574/583). This loop doesn't, and it never did — but tripling the window from 5s to 15s triples how long a `stop()` mid-setup leaves this loop spinning against a display ID that `stop()` already tore down (`virtualDisplay = nil` at line 421) before finally throwing into a dead call chain. Not a crash, just wasted work and log noise on exactly the unhealthy-machine path this PR is meant to help diagnose. A `if stopped { throw CancellationError() }` inside the loop would close it out now that the window is real enough to matter.

- **The branch predates the arrangement-persistence work on `main`** (PR #127, merged as `2324757`). `git merge-base pr-126 origin/main` is `3f13f9c`, three commits behind — before `DisplayArrangement`/`restoreOrigin`/`manageOrigin` existed. A trial merge doesn't textually conflict (different regions of `VirtualDisplay.swift`/`MacSender.swift`), but the commit-2 message's stated cost — "every existing user's Mac forgets the display's saved position once" from the productID bump — was written against the old serial-keyed-only world. On current `main`, placement is already tracked independently via `DisplayArrangement` keyed on `info.id`, which should make that one-time forgetting non-events for anyone whose receiver reports an id. Worth rebasing before the ship decision so the risk writeup in the PR body reflects what's actually on `main` now, rather than re-deriving it after merge.

## Minor

- `MacSender.swift:260`: `let ppi = info.kind == "iPad" ? 264.0 : 460.0` defaults to iPhone PPI whenever `kind` is unreported (older receivers omit `device`, so `kind` falls back to the literal string `"device"`). That's a no-worse-than-before fallback (old code used a phone-sized constant unconditionally too), but it means the exact users on old receiver builds don't get the corrected physical size that's the point of this change — the fix silently doesn't apply to them. Not blocking, just worth knowing the fallback population isn't covered.

- The three extra advertised modes in `Mac/VirtualDisplay.swift:55-63` are already flagged as a cost in #111 itself (System Settings will list them as selectable even though `selectHiDPIMode`'s enforcement loop fights back every 2s) — no new finding here, just confirming the tradeoff is accepted knowingly rather than overlooked.

- Commit 2's core premise — that a single non-standard EDID mode and/or the old productID caused the TV misclassification — is, per #111 and the PR body itself, unvalidated on any machine that actually reproduces it. Shipping it changes identity (and, before a rebase, the saved-position cost) for every user to test a hypothesis that hasn't been confirmed even once. That's a reasonable experiment to keep parked behind a flag or on this branch, but not something to merge to `main` on the strength of "the maintainer's machine still extends correctly."

BLACKIELF:
Validated the **poisoned display identity / product-ID bump** part of this PR on another affected Mac, using a strict single-variable harness.

### Environment and symptom

- MacBook Pro, M4 Pro
- macOS 26.6.1 (25G76)
- OpenDisplay 1.16.1
- iPad receiver over USB
- Mirror works; Extend consistently fails with `virtual display never appeared in SCShareableContent`

With commit 1's longer wait and diagnostics, the failure remained:

```text
inMirrorSet=false active=false online=false main=false
```

`CGVirtualDisplay.apply(_:)` returned `true`, but the assigned ID never appeared in `CGGetOnlineDisplayList` or `SCShareableContent`.

### Single-variable A/B result

I held the descriptor, dimensions, modes, and serial constant:

```text
productID=0x4F53 (current "OS") -> apply=true, but offline/inactive for the full 10 s sample
productID=0x4F44 (new "OD")     -> apply=true, online/active immediately and visible to SCK
```

As a second control, I applied the monitor-like physical size and mode ladder while keeping `productID=0x4F53`; it still remained offline. On this machine, the identity bump itself was the load-bearing change, not the extra modes/physical-size change.

I then rebuilt OpenDisplay with `productID=0x4F44`. Extend started normally at 2388x1668 @60 Hz; receiver telemetry reached about 58 fps over USB with 1-2 ms RTT.

### Suggested upstream handling

1. Treat `apply=true` + ID absent from `CGGetOnlineDisplayList` as a WindowServer registration/identity failure, before entering the ScreenCaptureKit poll.
2. Test membership in `CGGetOnlineDisplayList`, not `CGDisplayIsOnline(id) != 0`; as already noted on #217, an unknown ID can return `-1`, which incorrectly passes `!= 0`.
3. Prefer one bounded recreate with a persisted fresh identity generation when this state is detected. If a simple release-wide fix is preferred, this test validates the `0x4F53 -> 0x4F44` bump, with the known one-time cost of losing the saved display arrangement.
4. Consider separating the identity recovery from the speculative monitor-profile changes; the latter did not affect this reproduction.

Related report: #206. I have not classified Wi-Fi behavior from my ad-hoc build as part of this bug: the iPad's `_opensidecar._tcp` Bonjour service is visible at port 9000, and the local build uses a different bundle ID/signature, so Local Network authorization is a separate variable.

peetzweg:
Merged PR #288 now covers the bounded fresh-identity recovery validated in this draft, including bumping both serial/product identity and persisting the working offset. The online-display-list validation/longer surfacing diagnostics and the TV-classification experiments remain separate. Please split those remaining pieces from the now-overlapping identity recovery before rebasing onto current `main`.

PR #130, opened 2026-07-10
feat(android): add AndroidReceiver app

ayufan:
## What this does

Adds an **AndroidReceiver** app that lets an Android tablet act as a receiver for the Mac sender, using the same OpenDisplay receiver contract as the iOS app: Bonjour-compatible discovery, a length-prefixed TCP stream, H.264 Annex B video, and JSON control messages for input and liveness.

- NSD advertisement of `_opensidecar._tcp`, TCP listener on `:9000`
- `MediaCodec` decode straight to a `SurfaceView`, with the Mac cursor drawn as an overlay
- Tap, drag, and two-finger scroll input sent back to the Mac
- Native / Balanced / Smooth quality profiles and a latency/FPS overlay toggle

The receiver is based on the Android receiver from #121 by @chenyuhao0628 — authorship is retained on the introducing commit, and that fork's source code is licensed under GPL-3.0.

## Changes on top of the imported receiver

- **Full English UI** — the imported receiver was Chinese-only; all strings are extracted to resources with English as the default locale, and the Chinese launch, status, and settings translations are retained as `values-zh`
- **Pause-on-disconnect setting** — optionally pause the receiver UI when the Mac disconnects
- **Reconnect fix** — the TCP server now detects stale connections (keepalive, payload timestamps, per-client reader threads) so stream capture recovers cleanly after a reconnect
- **USB documentation** — how the Mac reaches the receiver over ADB port forwarding
- **README in English** with a full Chinese translation in `README.cn.md`

Works over WiFi out of the box, and over USB together with the Mac-side ADB support from https://github.com/peetzweg/opendisplay/pull/129.

## Testing

I've been happily using this for the last few hours on a Xiaomi Mi Pad 6. I observed latency of around 60-70ms.

## Related

- https://github.com/peetzweg/opendisplay/pull/129
- https://github.com/peetzweg/opendisplay/pull/121


peetzweg (commented):
Thanks for putting this together, and for cleaning up the #121 fork's licensing/i18n along the way — the receiver contract is followed closely and most of the wire-level plumbing (framing, hello, ping/pong, cursor echo, keyframe request) lines up with what the iOS receiver and `MacSender.swift` actually speak. There's one protocol-compatibility bug I'd want fixed before merging, plus a handful of things worth addressing after; overall this needs the frame-size issue fixed, then it's good to merge from my side.

**Blocking**

- `LengthPrefixedProtocol.java:14,33` hardcodes `MAX_FRAME_BYTES = 1 << 20` and `read()` throws (killing the connection) on any wire frame bigger than that. `MacSender.swift` has no equivalent cap on the send side, and its own comments acknowledge IDR frames are disproportionately large ("each one is a bitrate spike"). At "Best" quality (18 Mbps, native resolution) on a large panel, a keyframe can plausibly exceed 1 MiB — and keyframes are exactly what gets forced on every reconnect (`needsKeyframe = true` in `becomeReady`). That's a connect → oversized-keyframe → `IOException` → reconnect loop that never recovers, i.e. the app could be unusable at native quality on bigger screens. Either raise the cap well above any realistic keyframe size or drop it and rely on the existing per-message JSON/video disambiguation instead.

**Worth fixing**

- Scroll direction looks inverted relative to the iOS client. `ScrollGestureTracker.move()` (lines 34-35) negates the raw pixel delta before dividing by scale; the iOS equivalent (`OpenSidecarPhoneApp.swift:740-741`) sends `(t.x - lastPan.x)/scale` un-negated, and `MacSender.handleScroll` doesn't touch the sign either. Same two-finger drag should send the same-signed `dx`/`dy` from both receivers — right now it doesn't. The Android self-test (`ProtocolSelfTest.testScrollGestureTrackerProducesPixelDeltas`) locks in the negation as intentional, so this may have been tuned by feel on a real device, but it's worth confirming which direction is actually correct against the Mac and reconciling the two clients.
- `H264SurfaceDecoder.queueFrame` (line 60-61) silently drops a frame when `dequeueInputBuffer` has nothing free, with no call to `onDecoderNeedsKeyframe()` — unlike every other failure path in that class (SPS missing, `IllegalStateException`, decode error), which do request a keyframe. Since periodic IDRs are disabled on the Mac (`MaxKeyFrameInterval: 3600`), a single dropped frame here can leave the picture corrupted for up to a minute until the next scheduled keyframe, instead of resyncing immediately like the sibling failure paths do.
- No CI wiring and no working Gradle build. `.github/workflows/` is untouched by this PR, and `AndroidReceiver/build.gradle` / `settings.gradle` aren't actually exercised — there's no `gradlew` wrapper checked in, and `scripts/build_debug_apk.sh` bypasses Gradle entirely, invoking `aapt2`/`d8`/`javac` by hand with a hardcoded `/Applications/Android Studio.app/.../jbr` path. `tests/java/ProtocolSelfTest.java` is a nice set of unit tests but nothing runs it automatically. As it stands this module can silently bit-rot — nobody will notice a break until they try to build it by hand.
- `AnnexB.telemetryPrefix()` (line 21) parses the Mac's per-frame `cap`/`snd` timestamps but is never called outside its own unit test — the "latency overlay" only surfaces RTT and `inp50` piggybacked on the ping channel, not the true end-to-end capture-to-glass latency the Mac is already sending on every frame. Not wrong, just an unused hook and a smaller metric than what's actually available.
- `DisplayProfile`'s quality knob changes the advertised panel size in `hello` (`pixelsWide`/`pixelsHigh`), which drives virtual-display *creation* on the Mac side, so switching profiles mid-session triggers `reconfigure()` — a full virtual-display teardown/rebuild with the screen flash that comment explicitly calls out as something `switchTransport` was designed to avoid. The Mac's own analogous `StreamQuality` knob only rescales capture/encode and leaves the virtual display at native size. Worth deciding whether Android's quality profile should behave the same way (scale the stream, not the reported panel).

**Minor**

- `OpenDisplayServer.handleMacJson` (line 175) already parses control messages with `org.json.JSONObject`, but for `cursor`/`cursorImg` it re-parses the same string with `MacControlMessage.parse`'s hand-rolled regexes (line 31) instead of pulling the fields off the object it already has. Two independent JSON parsers for the same wire format is a bit of a maintenance trap.
- `install_debug_apk.sh` has Chinese-only error strings baked into the script itself, which reads oddly next to the "Full English UI" framing in the PR description — dev tooling only, so low impact.

ayufan:
Addressed the full review feedback and force-pushed a history rebased from `e04239d`, with the changes split into functional commits.

- Raised the inbound frame limit to 64 MiB while retaining an allocation guard, with coverage for frames over 1 MiB and oversized rejection.
- Drain/retry decoder output on input backpressure and request a keyframe when a frame must be dropped.
- Matched the iOS scroll-delta sign and removed the duplicate regex control-message parser in favor of the existing `JSONObject`.
- Wired the per-frame `cap`/`snd` telemetry into the overlay for median end-to-end and encode latency.
- Kept the advertised panel geometry native so quality changes no longer rebuild the virtual display; stream quality remains controlled by the Mac.
- Renamed the module to `Android/`.
- Added the Gradle wrapper, JUnit execution, and Android CI; removed `Android/scripts/` so Gradle is the only build/install path.

Verification: `./gradlew --no-daemon clean testDebugUnitTest assembleDebug lintDebug` — BUILD SUCCESSFUL.

PR #131, opened 2026-07-10
feat(mac): seamless transport switching between USB and WiFi

peetzweg:
Seamless transport switching: plugging the cable in while a device streams over WiFi migrates the live session onto USB, and unplugging fails over to WiFi when the device's service is visible — the picture freezes for about a second and resumes; display and window layout survive.

**Why:** USB has lower, steadier latency, so it should win whenever available — and losing the whole session (display torn down, windows reshuffled) just because a cable moved made plugging in/out disruptive rather than an upgrade.

**How:** new `MacSender.switchTransport(to:)` swaps only the socket under the live pipeline — virtual display, capture and encoder stay up; the receiver sees a fresh connection and resyncs with a keyframe. The controller decides when (usbmuxd attach → upgrade, detach → failover). This retires the old "no transport switching" policy, whose real justification was that switching used to mean destroying and recreating the session. Judgment calls: `usbDisabled` (from Disconnect clicks) now gates auto-connecting only, not migrating a session the user deliberately started; sessions remember their Bonjour service name because browse results routinely arrive without their TXT record, and after an unplug that name is the only remaining link between session and device row.

Also includes a standalone fix: virtual-display creation retries ~16s when the serial is still registered to a just-quit instance, instead of parking the session on "Failed".

Tested on device (iPad, USB + WiFi): upgrade, failover, both round trips, WiFi-started sessions taking the cable, single device row throughout, disconnect staying disconnected.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -284,6 +284,12 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         var vd: VirtualDisplay?
         for attempt in 0..<8 {
             if attempt > 0 { try await Task.sleep(for: .seconds(2)) }
+            // A Disconnect during the retry window tore the session down. Bail
+            // before creating/assigning the display: the serial the old display
+            // held is likely free now, so a late attempt would *succeed* and
+            // resurrect the very zombie this retry exists to avoid. (Mirrors the
+            // `if stopped` checks in the permission-poll loops above.)
+            if stopped { return }
             vd = await MainActor.run {
                 VirtualDisplay(name: displayName,
                                pointsWide: pointsWide, pointsHigh: pointsHigh,
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -623,11 +623,18 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         }
         connectionReady = false
         dialGeneration += 1   // a USB dial still in flight must not adopt
+        let generation = dialGeneration
         connection?.cancel()
         connection = nil
         pendingSends = 0
         queue.asyncAfter(deadline: .now() + 1.0) { [weak self] in
-            self?.connect()
+            // Generation-guarded so a switchTransport (or another reconnect)
+            // that landed in this 1s window supersedes this dial instead of
+            // racing it — otherwise the queued connect() re-dials the new
+            // transport, briefly running two live connections. (No bare
+            // self-rescheduling asyncAfter — the pattern banned in #76.)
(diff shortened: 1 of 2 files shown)


peetzweg (commented):
Reviewed both commits against the description. The migration design holds up: `switchTransport` swaps only the socket under the live pipeline, `usbDisabled` correctly gates auto-connect but not migration, disconnect opts out both transports, and the retry loop's failure path still lands on "Failed" rather than hanging. One bug should be fixed before merge, and there are three races worth knowing about.

**Blocking**

1. **The new ~16s retry loop in `setupExtend` never checks `stopped`** (`Mac/MacSender.swift:291-301`). The adjacent permission-poll loops do (`if stopped { return }` after the sleep, lines ~208/~239), and this loop needs it more: a Disconnect during the retry window doesn't cancel the task, and since the retry exists precisely because a stale serial is blocking creation, a late attempt is likely to *succeed* right after the old display frees the serial — assigning `virtualDisplay` on a stopped sender and resurrecting exactly the zombie display this commit is meant to fix. It can then hold the serial a replacement session is retrying against. One-line fix: check `stopped` after the sleep.

**Worth a follow-up (or a note in the code)**

2. **The upgrade path has a destructive fallback.** `activeSession(coveringUSB:)` matches only via `installIDByUDID` or usbmuxd-name == service-name. On a first-ever cable attach for a WiFi-started session — before lockdown resolves the name and before any install-id mapping exists — `autoConnect` falls into the plain connect branch and dials a *second* USB session; `dedupeSessions` then keeps the cable and tears down the WiFi twin. That's a full display teardown/rebuild, not a migration, contradicting "single device row throughout / no screen flash" for that case.

3. **`switchTransport` can race a queued `scheduleReconnect` dial** (`Mac/MacSender.swift:433-450` vs `623-625`). The 1s `asyncAfter { self?.connect() }` has no generation guard, and `connectTCP` doesn't check `dialGeneration`, so a reconnect queued just before the switch re-dials the new transport — two live connections briefly, receiver's replace-on-connect flaps. Pre-existing exposure via `forceReconnect`, but automatic unplug/replug will hit the window far more often. (Same asyncAfter-without-handle pattern we banned in #76.)

4. **Failover visibility is sampled only at the detach instant.** `failover()` runs solely from the usbmuxd detach callback; a Bonjour service appearing two seconds later — well inside the grace window — never triggers failover and the session dies. Matches the description's letter, but it's a sharp edge users will hit.

Minor: `session.onUSB`/`usbUDID` flip before the new link is confirmed, so a failed migration's Disconnect opts out a transport the session never reached; and the final retry attempt logs "retrying" when no retry follows.

Issue #17, opened 2026-06-10
macOS receiver — use another Mac as a display


peetzweg:
## Goal
Turn a spare MacBook (or any Mac) into an extra display for your main Mac, the same way iPhones and iPads work today.

## Why it's feasible
The receiver is intentionally thin: TCP listener → Annex B reassembly → `AVSampleBufferDisplayLayer` — all of which exist on macOS exactly as on iOS. The hello message already negotiates panel size and scale, so a Retina MacBook panel would get a correctly sized HiDPI virtual display automatically.

## Approach
- New macOS receiver target reusing the receiver pipeline (the Network/CoreMedia/AVFoundation code is platform-independent; only the UIKit view shell needs an AppKit/SwiftUI equivalent)
- Full-screen window (or true full screen space) rendering the stream
- Input back-channel: trackpad/keyboard events from the receiver Mac forwarded as pointer/key input — likely its own follow-up
- Transport unchanged: WiFi via Bonjour out of the box; wired Mac-to-Mac could use Thunderbolt/USB-C networking (`bridge` interface appears automatically when two Macs are cabled)

## Acceptance
A second Mac running the receiver app shows up in the sender's device list and works as an extended display at native resolution.

peetzweg:
Feasibility research outcome — short version: **yes, a Mac receiver is very doable, and it's less work than the iOS one was.**

- The receiver role uses only public APIs, all of which exist on macOS: `NWListener` + Bonjour for discovery, the length-prefixed TCP protocol, VideoToolbox decode, and `AVSampleBufferDisplayLayer` (available on macOS via AVKit/CoreMedia just like iOS).
- The sender doesn't care what's on the other end: it builds the virtual display from whatever panel size/scale the hello message announces. A Retina MacBook announcing its native panel @2x maps exactly onto the existing HiDPI virtual display path — zero sender changes expected.
- Input is *simpler* than iOS: a display-only Mac receiver needs no touch mapping at all. (Forwarding the receiving Mac's keyboard/trackpad back to the host — KVM-style — would be a separate follow-up.)
- The work is essentially one new macOS target embedding the existing receiver core: the networking/decode layer is UI-framework-light; only the view layer (currently UIKit) and the cursor-sprite drawing need AppKit equivalents.
- Caveats: the receiving Mac should suppress sleep while streaming, and fullscreen presentation (borderless window vs. real fullscreen) needs care so the menu bar/notch don't eat pixels.

[...]

PR #145, opened 2026-07-12
feat(mac): standalone OpenDisplay Receiver app, a spare Mac as a display

### LINKED ISSUE #82: Mac->Mac
Seems to be a usecase. Could replace airplay? At least someone mentioned that, did not know this is possible. But why not 🤔 

Feature requested here:

+ https://www.reddit.com/r/ProductivityApps/comments/1uelalu/comment/oubf86m/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

+ https://www.reddit.com/r/ProductivityApps/comments/1un1342/comment/ovjpu8q/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

peetzweg:
Closes #82 / #17. A spare Mac becomes a true extended display for another Mac, via a new standalone **OpenDisplay Receiver** app that runs on macOS 12 and up.

**Why:** Mac to Mac is the cheapest platform expansion (both halves stay on Apple frameworks), and unlike AirPlay to Mac it is a real persistent *extend*, not mirroring. The receiver is its own bundle on purpose: the sender is pinned to macOS 14 by its capture stack, receiving only needs a decoder and a window, so a separate app lets Macs from around 2015 serve as screens. Direction and the comparison with the alternative PR are in #241.

**How:** The receiver core moves from `iOS/PhoneReceiver.swift` to `Shared/StreamReceiver.swift` (platform specifics injected) and is reused unchanged by `MacReceiver/MacReceiver.swift`: Bonjour listener on :9000, hello announcing the primary screen's framebuffer, AppKit video window with the cursor echo, sleep suppression while receiving. New target `OpenSidecarMacReceiver` (`com.peetzweg.opensidecar.mac.receiver`, deployment target 12.0) compiles `MacReceiver/`, `Shared/` and the log/Sparkle files from `Mac/`. The sender app itself is unchanged apart from the cable and cursor work below; the earlier in-app role tab is gone. The deliberate call: one release tag ships both Mac DMGs with lockstep versions, each with its own Sparkle feed (`appcast.xml`, `appcast-receiver.xml`), rather than a second release track.

Sender-side improvements that came out of field testing:
- Cabled Macs: every dial first forbids WiFi so a Thunderbolt Bridge or Ethernet link is used when present, falling back instantly otherwise; the row says "Cable" (6a92fae). Plain USB-C between two Macs cannot work (host to host), documented in README and on the site.
- Cursor over WiFi stuttered behind video frames on the shared TCP socket: positions now go over a UDP side channel the receiver advertises in hello as `cursorPort` (additive, no pv bump, TCP fallback), which also benefits iPhone/iPad over WiFi (3262e37, PROTOCOL.md 6.3). After an external review pass the channel confirms delivery (`cursorAck`) before leaving TCP and one session sequence orders both paths (99082cb).
- A cable plugged in mid-session now wins: the receiver lists its addresses in hello (`addrs`, additive), and a sender on WiFi probes them with WiFi forbidden and migrates the live session onto the first non-WiFi path that answers, Thunderbolt Bridge, Ethernet, or plain USB-C host-to-host on recent macOS (23edb6e, PROTOCOL.md 6.4).

Release plumbing: second fastlane lane (`mac build_receiver_release`), `OpenDisplayReceiver.dmg` uploaded next to `OpenDisplay.dmg`, second appcast committed in the same step. Tests CI builds the receiver scheme so a macOS 13+ API sneaking into `Shared/` fails the PR. **Before the first release with this merged**: add the receiver bundle id's Developer ID profile to the match repo once (`bundle exec fastlane match developer_id`), otherwise the readonly match in CI fails.

Verified: loopback run of both apps (stream up at 3024x1900, UDP cursor channel negotiated, receiver quit ends the sender session cleanly), receiver builds at the 12.0 floor, 34 Mac unit tests, iOS target builds, site builds. Two Mac field test over WiFi done on the earlier in-app shape; the standalone app and the Cable path still to be confirmed on hardware. Display only for now: KVM-style input from the receiving Mac is #147.


peetzweg:
Real-hardware feedback round (2027c63): the sender cursor was invisible on the receiving Mac (connect-time sprite raced the video view's creation — receiver now buffers and replays it on attach), the receiving Mac's own pointer is blanked over the video, and the announced panel drops the menu-bar/notch strip (`safeAreaInsets.top`) so native full screen is pixel-1:1 instead of letterboxed on every side. Notch-less displays announce the full panel as before.

peetzweg:
Added (1b4700d): the performance overlay is now available on the Mac receiver too — PerfOverlay moved to Shared/ and pinned to the bottom of the video window, behind a "Performance overlay" toggle in the panel (same `showAnalytics` default as iOS).

clasalle-lila:
@peetzweg would love to help test. Is there a special way to compile the receiving end?

peetzweg:
You just need to compile the desktop version. Both is in the same desktop app. You just need to switch one Desktop app to being the receiver. 

Happy to get some user feedback on this as I don't have a "testflight" for the desktop app just yet.

But yeah this is a nice feture i can test well on my end on two macbooks. Worked pretty well. Did not wanted to yeet it in on that day. Will revisit this asap. :)

wantaekchoi:
Ran this branch on a Mac→Mac setup today (MacBook Pro M5 sender, iMac21,2 24"
receiver, wired, macOS 26.5.2 / 26.5) and hit one bug that makes the receiver
unusable out of the box, plus one gap that matters for the headless case this
PR is for.

## The receiver window is created 0 × titlebar and never recovers

The stream is fine — `Receiving 3360×1890`, ~30 fps, 18-20 ms end to end — but
nothing appears on the receiving Mac. The window exists and is on screen; it is
just zero-width:

```
#101 pid=1218 layer=0 onscreen=true bounds=325,116 0x0
```

A debug build with a log line after each step of `showWindow()` shows where it
goes:

```
after NSWindow(contentRect:)   (0.0, 0.0, 1519.0, 887.0)
after contentView =            (0.0, 0.0, 1519.0, 887.0)
after center()                 (325.0, 257.0, 1519.0, 887.0)
after contentAspectRatio =     (325.0, 257.0, 1519.0, 887.0)
after makeKeyAndOrderFront     (325.0, 1112.0, 0.0, 32.0)
```

`initialContentRect` is correct throughout. The collapse happens on the first
layout pass, which `makeKeyAndOrderFront` triggers.

Cause is in `ReceiverVideoView.init`. `OverlayHostingView` is an `NSHostingView`,
so by default it reports an intrinsic content size to Auto Layout, and it is
pinned to all four edges of the content view with required `equalTo`
constraints. The window ends up at whatever size the overlay reports rather than
the size it was created at.

I reproduced that in isolation, running the same init, contentView, center,
aspect-ratio and makeKeyAndOrderFront sequence four ways:

```
no overlay at all                                  1519x919   (no collapse)
NSHostingView, default sizingOptions, empty        0x32       (collapses)
NSHostingView, default sizingOptions, real content 148x68     (still wrong)
NSHostingView, sizingOptions = []                  1519x919   (no collapse)
```

The third row is the part I did not expect: zero is not special. With the HUD
switched on the window still comes out wrong, just at the HUD's own size instead
of vanishing, so this is not only a hidden-HUD problem.

One line fixes it:

```diff
 let overlay = OverlayHostingView(rootView: ReceiverPerfOverlay(receiver: receiver))
+overlay.sizingOptions = []
 overlay.translatesAutoresizingMaskIntoConstraints = false
```

`sizingOptions = []` flips `intrinsicContentSize` from a real value to
`NSView.noIntrinsicMetric`, so the edge constraints stop being able to drive the
window. After that the window comes up at 1519×887 and the extended desktop is
visible.

## A sleeping receiver display is never woken

`updateSleepAssertion` uses
`ProcessInfo.beginActivity(options: [.idleDisplaySleepDisabled, …])`, which
prevents future idle sleep but does not wake a display that is already asleep.

For the case this PR targets — a spare Mac with no keyboard or mouse — the
display will typically have slept by the time a sender connects, unless display
sleep is switched off in System Settings. The session then starts, frames
arrive, the receiver logs `Receiving …`, and the panel stays dark with no way to
wake it from the receiving side. That is what happened here before I found the
window bug, and the two together made it look like nothing was working at all.

`IOPMAssertionDeclareUserActivity(… kIOPMUserActiveLocal …)` when a session
starts would cover it. I have not written that one, so I am only reporting it.

Everything else worked as described: Bonjour discovery, the wire protocol
unchanged, no sender-side changes needed. Latency came out at 18-20 ms e2e over
a real two-machine wired link, which is not directly comparable to the ~14 ms p50
in your description (that was same-machine loopback) but lands in the same
ballpark.

Issue #123, opened 2026-07-09
iPad -> iPad


peetzweg:
Someone mentioned this in the comments of this article:

https://www.ifun.de/opendisplay-kostenlose-sidecar-alternative-aus-deutschland-283401/?utm_source=dlvr.it&utm_medium=twitter

peetzweg:
## Feasibility findings & implementation — PR #144

**What iPad → iPad can be on iPadOS:** the platform has no virtual-display primitive (no `CGVirtualDisplay` equivalent) and no event injection, so a true *extend* with touch-back is off the table. What it can be is **whole-screen mirroring, view-only** — and the only sanctioned way to capture beyond your own app on iOS is a **ReplayKit broadcast upload extension**.

**Design (implemented in #144):**
- The iOS app gains a sender role via a broadcast upload extension (`Broadcast/`, target `OpenSidecarBroadcast`). It speaks the exact wire the Mac sender speaks — low-latency VideoToolbox H.264, Annex B over framed TCP, telemetry prefix, `welcome`/`pong` control replies — so the **existing receiver app works unchanged** and the perf overlay/latency metrics keep working. Receiver `touch`/`scroll` messages are ignored (view-only).
- A new **“Send this screen”** sheet browses `_opensidecar._tcp` receivers (excluding itself), with `includePeerToPeer` on both ends so two iPads need no shared WiFi. The chosen target is handed to the extension via app-group defaults (`group.com.peetzweg.opensidecar.ios`) — the only channel the two processes share. Broadcasts start/stop only via the system picker / Control Center (no API to start one programmatically); connection loss ends the broadcast with a visible system alert instead of a dead recording indicator.
- iPhone works the same in both roles (iPhone→iPad, iPad→iPhone).

**Caveats / follow-ups:**
- **Provisioning:** the new App Groups capability means the first signed build must come from Xcode with the account signed in (profiles regenerate once); the TestFlight/release profile needs the capability too.
- **Rotation:** no explicit handling in v1 — if ReplayKit changes buffer dimensions on rotate the encoder rebuilds and the receiver resyncs off the fresh SPS/PPS; honoring the orientation attachment is a follow-up.
- **Overlap:** #122 (iPad→Mac) builds the same iPad-sender machinery and #82 (Mac→Mac) a receiver — whichever lands last should unify the sender pipeline rather than keep copies.

**Effort landed:** one new extension target + picker UI; zero receiver changes, zero wire-protocol changes.

PR #146, opened 2026-07-12
feat: use the Mac as a display for an iPad/iPhone (#122)

### LINKED ISSUE #122: iPad->Mac
Use mac as external display for the iPad 🤯

Someone mentioned this as a usecase as a comment in the original demo video:
https://www.youtube.com/watch?v=wyEUkMgH3zw

Interesting. Never tried using an external monitor on my ipad just yet. Will check it out.

peetzweg:
Closes #122.

Adds the reverse streaming direction: a device mirrors its screen into a window on the Mac (the existing direction is Mac → device).

**Why:** requested in #122 — people want their Mac to show their iPad, not only the other way around.

**How:** the Mac gains a receiver role (`MacReceiver`): an `NWListener` advertising a distinct Bonjour type (`_opensidecar-rev._tcp`) that decodes the *same* `[len][Annex-B]` wire the forward direction already speaks and renders each inbound device into its own `NSWindow`, rotation driven by a per-frame orientation tag. iOS gains a sender role via a ReplayKit **broadcast upload extension** (`OpenSidecarBroadcast`) — the only public whole-screen capture on iPadOS — reusing `MacSender`'s VideoToolbox H.264 + framed-TCP path with the roles inverted (device dials, Mac listens). App ↔ extension hand off the chosen Mac over an app group. One deliberate call: it's a **view-only mirror**, because iPadOS exposes no virtual-display or event-injection API — the Mac can show the screen but not extend or control it.

**Verification:** drove the Mac receiver end-to-end over loopback with a test sender reproducing the extension's wire behavior — handshake, 270 frames rendered in the window, mid-stream rotation, ping/pong all confirmed. Notes: the new App Groups entitlement needs a one-time provisioning-profile regen before a signed build; the broadcast extension itself can only be exercised on a real device (Control Center / system picker).

Overlaps the sibling `feat-ipad-ipad` (#123) and `feat-mac-mac` (#82) branches on `project.yml` and the iOS UI — whoever merges second should unify the one shared sender pipeline rather than keep copies.


peetzweg:
Closing this shape: the Mac-side receiver window here is superseded by the standalone OpenDisplay Receiver app that shipped in v1.18.0 (#145, decision in #241). The half of this branch still worth having is the iOS broadcast-upload sender for iPad to Mac; that should come back as a much smaller PR targeting the shipped receiver instead of carrying its own Mac window. #122 stays open to track it.

Issue #29, opened 2026-06-11
Multi-device: rotation triggers infinite cross-session rebuild loop (capture recovery rebuilds the display)


peetzweg:
## Symptom
With two devices connected, rotating one device can throw both sessions into an endless alternating rebuild loop — displays flip/flash every ~3s, devices appear to keep 'switching between horizontal and vertical'. Observed live 2026-06-11 ~21:31 with iPhone (USB) + iPad (USB) on `feat/multi-device` tip (now in main, v0.5.0).

## Root cause (confirmed from logs)
Destroying+creating a virtual display (the rotation rebuild) kills the **other** session's SCStream as an OS side effect:

```
21:31:43.828 reconfiguring for 2360x1640                  ← session A rebuilds its display
21:31:43.966 stream stopped with error: SCStreamErrorDomain Code=-3815
             "Failed to find any displays or windows to capture"   ← session B's stream dies
21:31:46.967 capture died — rebuilding pipeline           ← B's recovery…
21:31:46.968 reconfiguring for 1179x2556                  ← …does a FULL display rebuild
21:31:47.027 stream stopped … -3815                       ← which kills A's stream → loop
```

`scheduleCaptureRecovery` (MacSender) responds to a dead stream with `reconfigure(lastHello)` — a **full display destroy+create**. That recovery predates multi-device, where it was harmless: there was no second stream to kill. With N≥2 sessions, each rebuild kills the neighbor's stream and the recoveries ping-pong forever (~3s period = the recovery delay).

Two amplifiers observed in the same log:
- The mode-enforcement recovery (`@2x mode vanished — re-applying settings`) fires during every cycle; `display.apply(settings)` itself generates display-reconfiguration events, adding fuel.
- Each rebuild also re-triggers the other display's saved-mode restore fight.

## What already works (don't regress)
- Single rotation with 2 devices CAN recover cleanly (verified 20:57 same day: landscape→portrait→landscape, correct modes, no loop). The loop needs the timing where a rebuild lands while the neighbor's enforcement/recovery is mid-cycle — rotating the iPad (21:21, `reconfiguring for 1640x2360`) and then the iPhone seeded it.
- Mode enforcement itself (re-select @2x every 2s) fixed real bugs: stale-restored modes pillarboxing the framebuffer and 1x blurry relapses. Keep it; tame its recovery path.

## Fix plan (in order of importance)
[...]

PR #157, opened 2026-07-16
fix(mac): re-attach capture on stream death instead of full display rebuild (#29)

kdbhalala:
# fix(mac): re-attach capture on stream death instead of full display rebuild (#29)

Addresses issue #29 (rotation infinite rebuild loop) — fix-plan point 1.

## Root cause
`scheduleCaptureRecovery` responded to a dead `SCStream` with a **full
display destroy+create** (`reconfigure`). With ≥2 devices that rebuild kills
the *neighbor's* stream (OS side effect), whose recovery rebuilds *its*
display, killing the first — an endless ~3s ping-pong. Observed live
2026-06-11 (iPhone USB + iPad USB).

## Fix
A dead stream does NOT mean the display is gone. If our own virtual display
still exists, re-attach the capture to it (no destroy/create) — this alone
breaks the loop. Full `reconfigure` is now only the fallback when the display
is genuinely gone (e.g. display sleep tore it down).

- `scheduleCaptureRecovery`: check `CGDisplayBounds(vd.displayID)`; if live,
  `findSCDisplay` + `startCapture` at pixel res (`points*2 @2x`, matching
  `setupExtend`); on failure fall back to full rebuild.
- #9's `missingTicks` debounce already implements #29 point 2 (enforcement
  recovery no longer amplifies the loop).

## Not in this PR (follow-ups)
- #29 point 3 (serialize rebuilds across sessions with a shared lock) — needs
  a 2-device repro to validate; left separate to keep this PR minimal.
- Point 2's debounce lives in the #9 change.

## Verification
- Logic: the re-attach path is guarded by a non-null display bounds check.
- **Requires 2 devices (iPhone + iPad) to reproduce/verify the loop** — not
  runnable on a single device. Manual: 2 connected, rotate one then the other
  within ~10s, watch `/tmp/opensidecar-mac.log` for alternating
  `reconfiguring` / `-3815` / `capture died` (should be absent after fix).

References #29, #9, #26.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -479,11 +479,42 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         scheduleCaptureRecovery()
     }
 
-    /// Retry until capture is back (a rebuild during display sleep can fail).
+    /// Retry until capture is back. Per issue #29 fix-plan point 1: a dead
+    /// stream does NOT mean the display is gone. If our own virtual display
+    /// still exists, just re-attach the capture to it — rebuilding the display
+    /// (destroy+create) is what killed the NEIGHBOR's stream and ping-ponged
+    /// the infinite rebuild loop. Only do a full `reconfigure` when the display
+    /// is actually gone (e.g. display sleep tore it down).
     private func scheduleCaptureRecovery() {
         queue.asyncAfter(deadline: .now() + 3.0) { [weak self] in
             guard let self, !self.stopped, self.stream == nil,
                   let hello = self.lastHello else { return }
+            // Does our virtual display still exist? CGDisplayBounds returns a
+            // zero rect for an unknown id, so a non-zero bounds means it's live.
+            if let vd = self.virtualDisplay,
+               !CGDisplayBounds(vd.displayID).isNull {
+                Log.info("capture died — display still present, re-attaching capture only (#29)")
+                Task {
+                    do {
+                        let display = try await self.findSCDisplay(id: vd.displayID)
+                        // Capture at the display's pixel resolution (points ×2 @2x),
+                        // not SCDisplay.width (logical points) — matches setupExtend.
+                        let captureW = (Int(Double(vd.pointsWide * 2) * self.quality.scale)) & ~1
+                        let captureH = (Int(Double(vd.pointsHigh * 2) * self.quality.scale)) & ~1
+                        try await self.startCapture(display: display,
+                                                    pixelsWide: captureW, pixelsHigh: captureH)
+                        self.needsKeyframe = true
+                    } catch {
+                        Log.info("re-attach failed (\(error)) — falling back to full rebuild")
+                        await self.reconfigure(hello)
+                    }
+                    self.queue.async {
+                        if self.stream == nil { self.scheduleCaptureRecovery() }
+                    }
+                }
+                return
+            }
+            // Display genuinely gone — full rebuild (preserves old behavior).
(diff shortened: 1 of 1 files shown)


peetzweg:
Thanks, the core idea is right and it is the approach I want for #29: a dead `SCStream` does not imply the display is gone, and re-attaching instead of destroy+create is what breaks the cross-session ping-pong. `scheduleCaptureRecovery` is still untouched on main, so this still applies cleanly.

One real bug before it can go in. The guard never fails:

```swift
if let vd = self.virtualDisplay, !CGDisplayBounds(vd.displayID).isNull
```

`CGDisplayBounds` returns `CGRect.zero` for an unknown display id, not `CGRect.null`, and `CGRect.zero.isNull` is `false`. Verified:

```
zero.isNull = false | zero.isEmpty = true
bounds(invalid id 9999) = (0.0, 0.0, 0.0, 0.0) isNull = false
```

So the condition is always true. Two consequences: the "display genuinely gone, full rebuild" fallback is dead code, and the display-sleep case this function was originally written for now retries re-attach forever instead of rebuilding. The comment in the PR describes the intended behaviour correctly ("a non-zero bounds means it's live"), the predicate just does not match it. `.isEmpty` should do it, or check `width > 0 && height > 0` explicitly.

Two smaller things:

1. The `Task` inside the `queue.asyncAfter` block captures `self` strongly and reads `self.stream`, `self.quality` and `self.virtualDisplay` off `queue`. Worth confirming the re-attach path stays on `queue` for those reads, since the surrounding code is careful about that.
2. Verification is currently logic-only by your own note. Since the whole point is the 2-device loop, I would like an actual repro before merging: two devices connected, rotate one then the other within ~10s, and `/tmp/opensidecar-mac.log` free of the alternating `reconfiguring` / `-3815` / `capture died` pattern. I can run that here if you would rather hand it over.

Fix the guard and I will take another look.

peetzweg:
Verified live with two devices (iPhone on USB plus iPad on WiFi), which is the repro this PR said it needed.

**Before** (main, without this PR): rotating one device produced 19 full rebuilds and 24 virtual displays in 90 seconds, never converging. The virtual display counter climbed from 14 to 38.

**After** (this PR cherry-picked onto current main): 0 full rebuilds, 1 stream death, 1 re-attach, converged in about 3 seconds across 5 rotations. Displays created now tracks the rotation count instead of running away. The re-attach path fires exactly as designed:

```
[18:16:27.423] stream stopped with error: ... Code=-3815
[18:16:30.424] capture died — display still present, re-attaching capture only (#29)
[18:16:30.534] capture started: 1640x2360 display 123 mode extend
```

I pushed one fixup commit: the liveness gate tested `isNull` where it meant `isEmpty`. `CGDisplayBounds` returns `CGRectZero` for an unknown id, and `CGRect.isNull` is true only for the special `CGRect.null`, so `!bounds.isNull` read as "live" for a dead display too, which made the re-attach branch unconditional and the full-rebuild fallback unreachable. A genuine teardown still recovered, but only via the `catch` after `findSCDisplay` burned its ~5s of retries. Probed against CoreGraphics to be sure: bounds for a bogus id are `(0,0,0,0)`, `isNull` false, `isEmpty` true.

Two residuals, both outside this PR and left tracked separately: the neighbour stream still dies once per rotation and waits the hardcoded 3s before re-attaching (that is #29 fix-plan point 3), and virtual display origins are unstable across rotation, which turns out to be the visible symptom that remains once the loop is gone. Filing that second one as its own issue.

Thanks for this, it fixed the worst behaviour.

PR #162, opened 2026-07-20
perf: reduce lag spikes when scrolling or drawing fast

jlfwong:
## Summary

Serializes VideoToolbox H.264 encodes (`maxPendingEncodes = 1`) while keeping a small TCP send pipeline (`maxPendingSends = 3`). When either stage is saturated, ScreenCaptureKit frames are dropped **before** encode — never encode-then-discard. Pre-encode drops do **not** force keyframes, so the H.264 reference chain stays valid.

## Before this PR

**Encode times under load regularly pass 100ms**

image

https://github.com/user-attachments/assets/223da48a-459f-4535-b5a6-acd0db181b4e

## After this PR

**Encode times after this PR stay stable around 30ms**

image

https://github.com/user-attachments/assets/1fef65c4-d1f4-487a-b9a4-fbe781f93ca5

---

## Problem

`VTCompressionSessionEncodeFrame` returns immediately; the hardware encoder runs asynchronously. Under load, ScreenCaptureKit can deliver a new frame before the previous encode callback completes. VideoToolbox then runs **multiple encodes in parallel** inside the same session.

Profiling on full pipeline showed this hurts latency rather than helping throughput:

- Encode work **overlaps on the timeline** (parallel slices, not faster end-to-end output).
- Per-frame encode time **climbs through a burst** (roughly ~15 ms early in a burst → **100 ms+** on the tail).
- **cap→display p95 tracks the slowest in-flight encode**, so pen strokes and scrolling feel laggy even when average FPS looks fine.

Previously we had a single `drops` counter and treated encoder and network backpressure as one bucket, which made tuning and diagnosis harder.

---

## Solution

Two independent backpressure limits, checked at capture time:

| Limit | Value | Meaning |
|-------|-------|---------|
| `maxPendingEncodes` | **1** | Never pipeline hardware encodes; skip capture while an encode is in flight (**enc↓**). |
| `maxPendingSends` | **3** | Allow a small TCP pipeline so the link isn't idle between ACKs; skip capture when the send queue is full (**net↓**). |

Both paths call the same pre-encode drop helper:

- Drops are **invisible to the decoder** (reference chain intact → normal P-frame `n → n+2`).
- We do **not** force keyframes on encoder drops (avoids IDR pulsing / blockiness).
- We never waste VT work encoding a frame we'd discard on the network side.

---

## Test setup

| Component | Details |
|-----------|---------|
| **Mac** | MacBook Pro (Mac16,5), Apple **M4 Max**, 128 GB RAM |
| **macOS** | 15.1.1 (24B2091) |
| **iPad** | iPad Pro 12.9-inch (6th generation, **iPad14,5**) |
| **iPadOS** | 26.5.2 |

Repro: connect over USB, enable Performance overlay, draw quickly with Apple Pencil (or fast scroll/drag to load the encoder). Compare **enc↓** growth and cap→display latency before vs after.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -117,12 +117,30 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // multiple OpenDisplay monitors apart and persist their arrangement.
     private let displaySerial: UInt32
 
-    // Backpressure: outstanding sends. If the socket can't keep up we drop
-    // frames instead of queueing latency, then force a keyframe to resync.
-    // Kept tight: at 60fps each queued send is ~17ms of added latency.
+    // ── Encoder parallelism limiter (maxPendingEncodes = 1) ─────────────────
+    //
+    // VTCompressionSessionEncodeFrame returns immediately; the hardware H.264
+    // encoder runs asynchronously. If ScreenCaptureKit delivers the next frame
+    // before the previous encode callback fires, VideoToolbox will run multiple
+    // encodes in parallel inside the same session.
+    //
+    // Capping pendingEncodes at 1 enforces “latest frame wins” on the encoder:
+    // skip captures while an encode is in flight (enc drops), then feed the next
+    // fresh buffer when the callback clears the slot. The H.264 reference chain
+    // stays valid (pre-encode skip → normal P-frame n→n+2); we do NOT force
+    // keyframes on enc drops.
+    //
+    // Separate from network backpressure below (maxPendingSends): enc drops mean
+    // “encoder busy”; net drops mean “TCP send queue full”.
+    private var pendingEncodes = 0
+    private let maxPendingEncodes = 1
     private var pendingSends = 0
     private let maxPendingSends = 3
-    private var dropsThisWindow = 0
+    private let pipelineLock = NSLock()
+    private var dropsEncThisWindow = 0
+    private var dropsNetThisWindow = 0
+    private var dropsEncTotal = 0
+    private var dropsNetTotal = 0
     private var needsKeyframe = true
     private var connectionReady = false
     private var stopped = false
@@ -146,7 +164,7 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // Liveness: both sides ping every 2s; if nothing arrives for 5s the link
     // is half-open (e.g. usbmuxd accepted but the device is gone) — reconnect.
     private var lastReceived = Date()
-    private var dropsTotal = 0
+    private var dropsTotal: Int { dropsEncTotal + dropsNetTotal }
 
     // Local cursor echo: a cursor baked into the video carries the full
     // capture→encode→stream→display latency (~30ms perceived). Instead we
@@ -454,6 +472,9 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
             self.connection?.cancel()
             self.connection = nil
             self.pendingSends = 0
+            self.pipelineLock.lock()
+            self.pendingEncodes = 0
+            self.pipelineLock.unlock()
             self.connect()
         }
     }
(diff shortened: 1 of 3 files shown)


jlfwong:
This supercedes #119, and is related to the work for general iPad support in #4

peetzweg:
Reviewed the code and ran an independent A/B benchmark against main. This looks good, thank you for the thorough writeup.

## Code review

- The core design is right: drops happen at capture time, before `VTCompressionSessionEncodeFrame`, so the encoder state advances continuously and the reference chain stays valid. Notably, the old `needsKeyframe = true` on drop was already unnecessary (drops were pre-encode back then too), so removing it eliminates IDR pulsing at no correctness cost. That turns out to be the change with the biggest measurable effect (see below).
- Threading checks out: the SCStream sample handler and the NWConnection callbacks share the same serial queue (`sender.video`), so the counters need no lock there. `pipelineLock` covers exactly the one variable that needs it, `pendingEncodes`, which is decremented on VideoToolbox's callback thread. The synchronous failure path of `EncodeFrame` also releases the slot, so it can't leak.
- Two non-blocking nits: `shouldDropFrame(reason: String)` would read better as an enum, and the reconnect paths reset `pendingEncodes = 0` while a VT callback may still be outstanding, so a late callback can briefly undercount and allow two overlapping encodes right after reconnect. Both are cosmetic and self-correcting, fine to leave as is.

## Independent benchmark (MacBook + iPhone 15 Pro, WiFi)

Same iOS build throughout, Mac app swapped between main and this branch. Load: a fullscreen scrolling checkerboard on the virtual display (fast-scroll workload, 95s, run A-B-A) plus a full-frame noise load (encoder worst case, 65s). Medians of the 5s stats windows:

| | scroll: main | scroll: this PR (2 runs) | noise: main | noise: this PR |
|---|---|---|---|---|
| encode p50 | 8 ms | 8 / 7.5 ms | 7 ms | 7 ms |
| e2e p50 | 17 ms | 17 / 16 ms | 16 ms | 12 ms |
| **e2e p95** | **157 ms** | **70 / 70 ms** | **117 ms** | **13 ms** |
| fps | 56 | 58 / 59.5 | 53.5 | 53 |
| drops per run | 26 | 15 / 7 | 15 | 2 |

Two observations:

1. The tail latency win reproduces cleanly: e2e p95 roughly halves, confirmed A-B-A (70 -> 157 -> 70 ms), with fps, bitrate and median latency unchanged. The mechanism matches your analysis, except the dominant cause on my setup is the forced keyframes: main forced an IDR on every one of its 26 drops and everything queued behind those bursts.
2. The encode pileup itself does not reproduce at iPhone resolution: encode p50 stays pinned at 7-8 ms on both builds, well inside the 60fps frame budget, so encodes never overlapped in the first place. Your iPad 12.9 target pushes much larger frames, which explains the difference. On smaller targets `maxPendingEncodes = 1` is dormant insurance, which is fine.

## Compatibility

The wire change is purely additive: the ping keeps `drops` with the same meaning, old receivers ignore the new keys, and new receivers fall back to `drops` when `encDrops` is absent. Per COMPATIBILITY.md this needs no protocol version bump and no update forcing, so it is safe for the whole installed base, including users still on 1.0.

Merging.

mad0x20wizard:
You guys are the best. I installed OpenDisplay only a couple of days ago and had to find that the native speed/quality was practical unusable on my setup (M1 MacBook and iPad). Now the performance counters still flash a lot of red between the yellow baseline, but the subjective performance is nearly as smooth as with Apple's native Sidecar. Thank you very much for developing and supporting this project! 😁

peetzweg:
@mad0x20wizard thanks! It's live in testflight built and coming to App Store version asap. A Glad you enjoy the app! Ein AppStore review oder auch spende ist immer willkommen 🙌:)

mad0x20wizard:
Ah, du kommst ja aus Berlin. Habe schon gespendet, schreibe bei Gelegenheit eine Review. ⭐️

peetzweg:
Mega danke dir @mad0x20wizard 🙏🙌

PR #163, opened 2026-07-20
Apple Pencil support w/ pressure/tilt/proximity

jlfwong:
# Apple Pencil input over the control channel

Adds end-to-end Apple Pencil support to OpenDisplay: the iPad captures hover and stroke data, sends it over the existing JSON control channel, and the Mac injects native tablet events (pressure, tilt, proximity) onto the virtual display.

Barrel roll support will be introduced later, I'm waiting for an Apple Pencil Pro in the mail :)

Finger touch behavior is unchanged — tap/drag/scroll still go through the existing `touch` and `scroll` messages.

## Pressure Support Demo

https://github.com/user-attachments/assets/e8a81c03-d1d6-4c21-9dcd-5454f3c4a7d0

## Tilt Support Demo

https://github.com/user-attachments/assets/5036e39c-5e91-4329-bac8-65c01c3f210c

### Wire protocol

Two new control message types:

**`pencil`** — stroke and hover updates
```
{
  "type": "pencil",
  "phase": "hover" | "down" | "move" | "up",
  "x": 0.0, "y": 0.0,
  "pressure": 0.0,
  "azimuth": 0.0,
  "altitude": 1.57,
  "rotation": 0,
  "t": 1234567890.0
}
```
- `x` / `y`: normalized video coordinates (top-left origin), same as finger touch.
- `azimuth`, `altitude`: radians from UIKit.
- `rotation`: always `0` for now (see **Out of scope** below).
- Optional `t` timestamp for Mac-side input latency stats.

**`proximity`** — pen enter/exit range
```
{
  "type": "proximity",
  "entering": true,
  "x": 0.0,
  "y": 0.0
}
```

### Mac — injection (`InputInjector`)

- Routes `pencil` and `proximity` messages from `MacSender`.
- Posts **`CGEventType.tabletProximity`** on enter/exit with a synthetic OpenDisplay tablet identity (vendor `0x0D15`, device `424242`, Grip Pen `0x0802`) so apps like Krita/Qt register a tablet device.
- **Pressure strokes** → `CGEvent` tablet-point events (`mouseEventSubtype = tabletPoint`) with pressure, tilt, and device ID.
- **Zero-pressure taps** → plain left mouse down/up (for clicking UI elements).
- **Hover** → tablet-point `mouseMoved` with tilt fields (enables tilt preview in browsers and drawing apps).
- **Off-screen drag recovery**: if hover resumes while Mac still thinks the button is down, releases the stale down before posting hover (fixes stuck mouse button after dragging off-screen).

## Behavior summary

| Input | Wire | Mac injection |
|-------|------|---------------|
| Finger tap/drag | `touch` | Mouse events (unchanged) |
| Finger two-finger pan | `scroll` | Scroll wheel (unchanged) |
| Pencil hover | `pencil` phase=`hover` + `proximity` enter | Tablet proximity + tablet hover with tilt |
| Pencil tap | `pencil` down/up + `proximity` | Mouse click |
| Pencil stroke | `pencil` down/move/up + `proximity` | Tablet down/drag/up with pressure + tilt |

## Testing

Manual testing on iPad Pro + **Apple Pencil (non-Pro)** over USB:

- [x] Pencil hover moves cursor and reports tilt in browsers
- [x] Pressure-sensitive strokes in Krita
- [x] Zero-pressure tap clicks buttons / UI
- [x] Finger touch, drag, and scroll still work
- [x] Hover cursor resumes after lifting pen post-stroke
- [x] Off-screen drag no longer leaves mouse button stuck on hover resume

**Not tested** (no hardware): Apple Pencil Pro barrel roll / `rollAngle`.

## Deferred: Barrel roll (Apple Pencil Pro only)

Regular Apple Pencil does not report `UITouch.rollAngle` (always `0`). Barrel roll requires **Apple Pencil Pro** + iPadOS 17.5+.

Rather than land untested Art Pen device-ID hacks and rotation conversion, this PR:

- Sends `"rotation": 0` on the wire
- Injects `tabletEventRotation = 0` on Mac
- Leaves `TODO` comments at the iOS capture, wire, and Mac injection sites


jlfwong:
This is the next PR for #4

peetzweg:
Picked up a borrowed apple pencil today! Will get to testing this asap. 🙏

peetzweg:
This is really nice work, thanks for building it. The Mac side tablet plumbing is more careful than I expected, and I like that it's fully additive: no Pencil, no behavior change. Also thanks for the demo videos in the description, being able to actually see the pressure and tilt working made this much easier to review.

Heads up that I can't test this on hardware myself, more on that below. So this is from reading it closely plus a build check.

**Needs a deployment target bump**

`UIHoverGestureRecognizer.azimuthAngle(in:)` and `.altitudeAngle` are iOS 16.4, not 16.0, so this no longer compiles now that #178 lowered us to 16.0. Could you add `deploymentTarget: "16.4"` to project.yml as part of this PR? I'd rather bump the floor than add an availability guard, since every device that can run iOS 16 can update to 16.7.x, so 16.4 costs us nobody. I checked and the bump alone is enough, no code changes needed. The stroke path is fine either way since `UITouch.azimuthAngle(in:)` is iOS 9.1, and hover needs an M2 iPad regardless.

**The entry-level Apple Pencil (USB-C) probably takes the wrong path**

This is the one I'd most like your read on. The USB-C Pencil has no pressure sensor, so `touch.force` never varies from about 0. If that's right, every stroke with it hits the `pressure < 0.01` branch in `handlePencil` and sets `pencilTapMode`, which posts a plain `leftMouseDown`, then tablet drags via the `move` case, then `leftMouseUp`. So mouse and tablet subtypes mix mid-stroke on what should be a normal stroke.

Since `pencilTapMode` also never gets cleared once the pen starts moving, I think this affects any light stroke on a pressure-capable Pencil too, just less often.

Would checking `touch.maximumPossibleForce == 0` to detect a non-pressure Pencil work better than inferring it from a single zero sample? Then tap vs stroke could stay a pressure question only on hardware that actually reports pressure.

**New iPad plus old Mac makes the Pencil go dead**

Today a `.pencil` touch falls through the generic `touch` path and works as a mouse. After this PR iOS only sends `pencil`/`proximity`, so an old Mac hits the `default:` case and the Pencil does nothing at all. That's the common skew direction, since the App Store auto-updates the iOS app while the Mac app waits on Sparkle.

Could you gate it on the Mac's protocol version? We already parse `pv` in `PhoneReceiver` at the `welcome` case and then discard it. Storing it, bumping `WireProtocol.version` to 3, and falling back to the legacy `touch` path for older Macs would keep Pencil-as-mouse working instead of dead. The slightly awkward part is that `pv` arrives on the welcome message after the input engine is wired up in `makeUIView`, so it probably needs a live read rather than a value captured at setup.

It also spams our Mac log a few hundred writes a second during a stroke, but that part is our bug in the `default:` case and I'll fix it separately.

**Palm rejection, could you check this one?**

`routeTouches` forwards fingers and pencil at the same time, and the finger gate only counts fingers, so a hand resting on the glass still gets forwarded. One resting contact point means the Mac gets a real `leftMouseDown` at the palm while the pen posts tablet drags elsewhere. Does that hold up if you rest your hand while drawing in Krita? If not, suppressing finger forwarding while a pencil touch is active would probably cover it.

**Smaller things**

- The `#available(iOS 17.0, *)` around `.indirectPointer` can go, it's iOS 13.4. It was harmless while we required 17.0 since it was always true, but at a 16.x target it evaluates false, so `isFinger` returns false for pointer touches. Do you know offhand whether trackpad or mouse clicks on the iPad actually come through `touchesBegan` here? If they do, they'd now get dropped by both branches.
- `case .pencil, .stylus` is redundant, `UITouchTypeStylus` is a deprecated alias for the same value.
- Not for this PR: the finger path sends a predicted touch to hide a frame of latency and the pencil path doesn't. That's likely where you'd feel it most while drawing.

**On testing**

I did want to try this myself. I borrowed an Apple Pencil, but it turned out to be a 1st generation one, and pairing that with my iPad needs the USB-C to Apple Pencil adapter, which I don't have. So I can't get hands on it right now, which is a shame.

Best I can do is merge once the target bump is in and push a TestFlight build to get real reports back from people who have the hardware. Happy to go that route, I'd rather get this in front of users than sit on it.

**Website**

I'd love to feature this on opendisplay.app. Would you be up for a short clip? And could you sanity check this table before I put it on the site, since you have the hardware and I don't:

| Apple Pencil | Pressure | Tilt | Hover |
|---|---|---|---|
| Pencil Pro | yes | yes | yes |
| 2nd gen | yes | yes | yes |
| 1st gen | yes | yes | no |
| USB-C | no | yes | yes |

Plus: hover needs a hover-capable iPad (M2 iPad Pro or Air and later, iPad mini A17 Pro), and barrel roll is not in this PR.

Nothing structural here, the design is right. Great addition, let's get it landed.

DrNeroCF:
Got this working on my 1st gen pro, and it’s the first time I’m seeing Sidecar-like Pencil support on my desktop, truly incredible!

Palm rejection is definitely an issue, though my perfect setup would be how Jump Desktop handles pencil+touch, where touch moves the cursor like a trackpad while the pencil inputs directly.

tapMoveThreshold does mess with the beginning of a drawn stroke, though.  I’m not sure if there’s a perfect fix for that, though.  I love the responsiveness when that’s disabled, but double tapping does becomes difficult…  Buffering the inputs until that threshold is crossed should draw the line correctly (as long as the program can keep up), but still creates some lag for stroke start.  Hm, I wonder if a timer could distinguish between the first tap of a double tap and a drag more quickly?  Ideally, that first down would cause an immediate response, then move suppressed for the least time possible.

Apologies for the backseat designing, I really need to learn some swift, apparently!

DrNeroCF:
Quick issue I noticed, in OpenSidecarPhoneApp, emitPencil is called at line 1084, and contains the same data as the last touch in coalescedTouches, so the data was coming in out of order and repeated, causing odd behavior when scribbling quickly.  The first move after down seems to contain some bad pressure data, but lines are more accurate overall using only coalescedTouches.

jlfwong:
> tapMoveThreshold does mess with the beginning of a drawn stroke, though. I’m not sure if there’s a perfect fix for that, though. I love the responsiveness when that’s disabled, but double tapping does becomes difficult… Buffering the inputs until that threshold is crossed should draw the line correctly (as long as the program can keep up), but still creates some lag for stroke start. Hm, I wonder if a timer could distinguish between the first tap of a double tap and a drag more quickly? Ideally, that first down would cause an immediate response, then move suppressed for the least time possible.

Was just testing this on my iPad yesterday during a drawing session, and I agree this is annoying and feels laggy and bad! Will try to fix this

jlfwong:
@peetzweg okay, made the changes as requested.

It looks like there's a conflict on `project.yml`. Is that something you can handle @peetzweg?

I attempted to fix the initial stroke delay while also making double clicks easier. Can you test it out @DrNeroCF?

## Commit 1: Fix Apple Pencil input for iOS 16.4 and mixed Mac versions (`f6632f2`)

### Problem
Several issues were blocking reliable Apple Pencil use: USB-C Pencil taps were routed through plain mouse events, stroke start was delayed, old Mac builds couldn't understand pencil wire messages, palm touches interfered with pen input, and some APIs were gated behind iOS 17 unnecessarily.

### Changes

**Mac (`InputInjector.swift`)**
- Always inject pencil as **tablet events** (`tabletPoint` subtype with pressure/tilt) for down, move, up, and hover.
- Removed `pencilTapMode` and the `postMouse` path that inferred taps from zero pressure — that path caused USB-C Pencil bugs when mixed with tablet events.

**iOS (`OpenSidecarPhoneApp.swift`)**
- Send **`down` immediately** on pen contact (removed delayed-down / movement threshold logic that added stroke-start lag).
- **Palm rejection**: suppress finger touches while the pen is active (`hasActivePen`).
- Handle **indirect pointer / hover** without an iOS 17 availability guard (API exists from iOS 13.4+).
- Remove redundant `.stylus` touch type handling (alias of `.pencil`).

**Wire protocol (`Protocol.swift`, `PhoneReceiver.swift`)**
- Bump protocol version to **3**; add `pencilWireVersion = 3`.
- iPad reads Mac protocol version from `welcome` and sets `macSupportsPencilWire`.
- When connected to an old Mac (`pv < 3`), pencil input **falls back to legacy `touch` events** so mixed-version pairs still work.

**Build (`project.yml`)**
- Lower iOS deployment target to **16.4** (needed for `UIHoverGestureRecognizer` azimuth/altitude APIs).

---

## Commit 2: Synthesize pencil double-click clickState on Mac (`84b3352`)

### Problem
After sending pen `down` immediately on contact, **double-tap was hard to execute**. Synthetic tablet events always had `mouseEventClickState = 1`, so macOS never saw a second click in a double-click sequence. The Window Server assigns click state for hardware mouse events, but not for our posted tablet events.

### Changes

**Mac (`InputInjector.swift`) — pencil only; finger/touch path unchanged**
- Track pen click sequences with optional `PenClickSession` / `PenCompletedClick` state (no sentinel values).
- On pencil **down**: compute `mouseEventClickState` (1, 2, 3, …) from timing and location vs. the previous completed click.
- On pencil **up**: emit matching click state; extend the multi-click chain only if **down→up displacement** is within the system threshold (a drag resets the chain).
- Read thresholds from the user's System Settings:
  - **Interval**: `NSEvent.doubleClickInterval` (public API)
  - **Distance**: AppKit's `NSDoubleClickDistance()` (same value the Window Server uses; loaded at runtime since it's not exposed in Swift)

peetzweg:
@jlfwong Merged main in, so the `project.yml` conflict is gone. I also pushed two commits straight to your fork. Hope that's fine. We are now on 16.4 which should be fine. Everyone who has access to 16.0 like before should be easily able to upgrade. So the project.yaml issue is gone.

@DrNeroCF thanks for reporting this can you give a try again? Is it better now?

If all good, let's merge and shit it 🔥

peetzweg:
@jlfwong Merged main in, so the `project.yml` conflict is gone. I also pushed two commits straight to your branch since maintainer edits are on. Heads up, and push back if you disagree with either.

**`efc3fe4` fix(ios): send pencil samples in the order UIKit recorded them**

@DrNeroCF was right, thanks for catching it. `coalescedTouches` already ends with `touch` itself, so `emitPen` sent the newest sample ahead of its own history and then repeated it. The `where c !== touch` guard never matched, because UIKit puts copies in the array.

Rather than just reorder it, I collapsed all three call sites (finger, `sendPencilAsTouch`, `emitPen`) onto one `UIEvent.samples(for:)` helper. Two of them were already correct, so this was copies drifting apart.

**`e016dc7` fix(mac): normalize pencil tilt into the CGEvent unit range**

`deriveTilt` passed `pi/2 - altitude` straight through, so a flat pen read 1.57 where `kCGTabletEventTiltX/Y` want -1...1. Krita clamps, which is why it survived testing.

Deployment target stays at 16.4 as discussed.

I ran a finger-only pass on my iPad against this build, since `routeTouches` restructured all four touch handlers and that part affects everyone: tap, drag, two-finger scroll, second finger joining mid-press, and menu open/close all behave. CI is green too.

@DrNeroCF could you pull the branch and check fast scribbling? You are the only one of us with the hardware, and I would like your read before this goes to TestFlight.

After that I will merge and push a TestFlight build. Filing follow-ups for palm rejection swallowing the finger lift, `InputInjector` keeping pen state across a disconnect, the missing predicted touch on the pencil path, and barrel roll.

jlfwong:
@peetzweg ship it!

re: video recording and feature description for website, would rather do that after landing keyboard shortcut sidebar & multi-touch gestures, but if you'd like to do it right away for the limited set, go ahead

Issue #72, opened 2026-06-28
What is the oldest iOS version we can support?


peetzweg:
Had a few people want to give their old iPad a second live. This is a huge usecase for this app. The older the iOS version we can support the better. iPads become dead paper weight at some points so this would be a safer for a lot of these.

Afaik we currenlty can do iOS17. We need to research what is the oldest version of iOS and what the iPad support is, like what is the oldest iPad we can support. We need a proper documentation and support table for this to make good decisions here and communicate this as a proper feature of the app.

Requests:
+ https://www.reddit.com/r/ProductivityApps/comments/1uelalu/comment/ou6jnld/

Kitround:
Incredible project, i'm using a buggy Yam Display because it's the only thing available in iOS 9.3.5 for a old iPad Mini v1. Will be a blessing to have a support on it.

peetzweg:
## Research: how low can the iOS deployment target go?

I traced every iOS-version-gated API in the receiver (`iOS/PhoneReceiver.swift`, `OpenSidecarPhoneApp.swift`, `MetalVideoRenderer.swift`). Short version: **the streaming pipeline is not the constraint — the SwiftUI UI layer is.** We're pinned at `deploymentTarget: 17.0` today (`project.yml`) almost entirely by *one* API used four times, and the realistic floor without a rewrite is **iOS 15**.

### What actually forces iOS 17 today

The whole core pipeline already supports very old iOS:

| Subsystem | Code | Available since |
|---|---|---|
| Transport (`NWListener`/`NWConnection`, Bonjour via `NWListener.Service`) | `PhoneReceiver` | **iOS 12** |
| H.264 decode + display (`AVSampleBufferDisplayLayer`, `CMVideoFormatDescriptionCreateFromH264ParameterSets`, `VTDecompressionSession`) | `PhoneReceiver` | iOS 8 |
| Metal path (`CAMetalLayer`, `addPresentedHandler`) | `MetalVideoRenderer` | iOS 10.3 |
| SwiftUI app lifecycle (`@main App`, `WindowGroup`, `@StateObject`, `@AppStorage`) | `OpenSidecarPhoneApp` | iOS 14 |

The only things dragging us up to 17:

- **iOS 17 (the real blocker):** the two-parameter `.onChange(of:) { old, new in }` — used 4× (lines 79, 95, 99, 538). Trivial to backport to the single-param `.onChange(of:perform:)` (iOS 14+).
[...]

PR #178, opened 2026-07-28
feat(ios): lower deployment target to iOS 16

patrickalessik:
Closes part of #72 — the iOS 16 step.

## What this does

Lowers `OpenSidecariOS` from `deploymentTarget: 17.0` to `16.0`, which revives the **iPad 5 (2017), iPad Pro 1st gen, iPhone 8 and iPhone X** — the population Apple left on the 16.7.x line. Three people in #72 (@Hugh0306, @Keox35, @doccstat) are on 16.7.16 specifically and have volunteered to test.

## Why it's this small

The research in #72 holds up: **the streaming pipeline was never the constraint.** `Network.framework` is iOS 12, `AVSampleBufferDisplayLayer`/`VideoToolbox` older still, the Metal path 10.3. The entire iOS 17 floor came from four two-parameter `onChange(of:) { old, new in }` closures in `iOS/OpenSidecarPhoneApp.swift` — and all four discarded the old value, so each is a 1:1 rewrite to the single-parameter form (iOS 14+):

| Site | Value observed |
|---|---|
| `:94` | `geo.size` (rotation → `setOrientation`) |
| `:129` | `scenePhase` |
| `:159` | `receiver.connected` (retires the onboarding hint) |
| `:599` | `deviceName` |

**No transport, decoder or renderer changes.**

### No availability shim needed

The single-parameter `onChange` is deprecated on 17+, so I expected to need an `#available`-branching helper to keep the build warning-free. It turns out the compiler only emits that deprecation warning when the deployment target is itself ≥ 17 — at 16.0 it's silent. I verified this both ways before dropping the shim, so this is the whole change rather than the change plus scaffolding.

## Verification

Compiled against the iOS 26.5 SDK (Xcode 26.6):

- Type-check at `arm64-apple-ios16.0` — **0 errors, 0 warnings**
- Optimized whole-module codegen (`-O -wmo`) — **0 errors, 0 warnings**, `minos 16.0` in the resulting object
- No `#available` / `systemVersion` gates exist anywhere in `iOS/` or `Shared/`, so the compiler's availability check is complete coverage rather than partial
- SDK `MinimumDeploymentTarget` is 12.0, so Xcode 26 has no objection to a 16.0 target

**Not yet smoke-tested on a real 16.x device** — I don't have one. This is the thing worth confirming before release; TestFlight's own minimum is iOS 16.0, so the 16.7.16 testers in #72 can be reached directly by a beta build.

## Docs

Corrects the stale `iPadOS 17+` / `iOS 17+` strings in `src/App.tsx` (hero meta line) and `index.html` (schema.org `operatingSystem`), and documents the requirement in the README's iPhone app section, naming the devices it revives.

## On the performance caveat from #72

The A8/A8X worry in that thread applies to the **iOS 15** tier, not this one. The oldest devices unlocked here are A9/A9X (iPad 5, iPad Pro 1), which handle 2048×1536 hardware H.264 decode comfortably. No adaptive quality cap looks necessary for 16.

## What's left for iOS 15 (not in this PR)

For reference, I ran the same type-check at `ios15.0`: **30 errors**, from `NavigationStack`, `ViewThatFits`, `LabeledContent`, `persistentSystemOverlays`, and the `Layout`/`ProposedViewSize` protocol behind `FlowLayout` — matching the manual audit in #72. That's a genuine SwiftUI backport rather than a one-line change, so it's left as a separate step. Note it would not help @bwagner's iPad Air 2 on 15.8.8 reach a build *via TestFlight*, since TestFlight itself requires 16.0 — that tier needs a sideloaded build either way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ opens with a plain double-click on macOS 14+ — no Gatekeeper warning. Open the
 
 ### iPhone app
 
+Needs **iOS / iPadOS 16 or newer** — including the 16.7.x line, which is where
+Apple left the iPad 5 (2017), the iPad Pro 1st gen, the iPhone 8 and the
+iPhone X. If your iPad can't be updated past 16.7, it can still be a second
+display ([#72](https://github.com/peetzweg/opendisplay/issues/72)).
+
 - **TestFlight** (recommended): join the public beta at
   [testflight.apple.com/join/3NYaY11c](https://testflight.apple.com/join/3NYaY11c).
 - **Build from source**: open the project in Xcode, select your free Apple ID
--- a/iOS/OpenSidecarPhoneApp.swift
+++ b/iOS/OpenSidecarPhoneApp.swift
@@ -91,7 +91,7 @@ struct ReceiverScreen: View {
                 }
             }
             .onAppear { model.receiver.setOrientation(portrait: geo.size.height > geo.size.width) }
-            .onChange(of: geo.size) { _, size in
+            .onChange(of: geo.size) { size in
                 model.receiver.setOrientation(portrait: size.height > size.width)
             }
             .sheet(isPresented: $showOnboarding) {
@@ -126,7 +126,7 @@ struct ReceiverScreen: View {
         .onReceive(NotificationCenter.default.publisher(for: .deviceDidShake)) { _ in
             showSettings = true
         }
-        .onChange(of: scenePhase) { _, phase in
+        .onChange(of: scenePhase) { phase in
             Log.info("scenePhase -> \(String(describing: phase))")
             switch phase {
             case .active: model.sceneDidActivate()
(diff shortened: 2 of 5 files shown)


patrickalessik:
Tested on real hardware — this now has the device confirmation the description was missing.

**iPad Pro 12.9" 1st gen (`iPad6,8`, A9X) running iPadOS 16.7.16**, sideloaded from this branch with a free personal team (empty entitlements file, so no paid capabilities needed). The app builds, installs, launches, and **streams** — the Mac desktop shows up on the iPad over USB, with video decoding fine on 2015 silicon. That's the exact version @Hugh0306, @Keox35 and @doccstat asked for in #72.

Also completed the full build I couldn't run earlier:

```
xcodebuild -scheme OpenSidecariOS -destination 'generic/platform=iOS' \
  CODE_SIGNING_ALLOWED=NO build
** BUILD SUCCEEDED **
```

0 errors, 0 warnings (the only output is the routine `appintentsmetadataprocessor` "No AppIntents.framework dependency found" note, which the iOS 17 build emits too). The produced app declares `MinimumOSVersion 16.0` in its `Info.plist`, with `minos 16.0` in the binary's `LC_BUILD_VERSION`.

One note for anyone else sideloading to test: `PRODUCT_BUNDLE_IDENTIFIER` has to be changed to your own, since `com.peetzweg.opensidecar.ios` can't be registered to another team. That's a local-only change and isn't part of this PR.

patrickalessik:
@peetzweg please merge and update app 🙏

Issue #142, opened 2026-07-12
Mac: virtual display created but never appears in SCShareableContent (Code=3) — no capture / purple indicator; CGDisplayStream fallback works


hmachine1999:
## Summary

On a **Mac mini (Apple Silicon) + single external 4K display** running **macOS 15.7.x**, OpenDisplay can:

1. Discover and connect to the iPad over Wi‑Fi (and hello is received)
2. Successfully create a `CGVirtualDisplay`

…but then fails with:

```text
sender failed to start: Error Domain=MacSender Code=3
"virtual display never appeared in SCShareableContent"
```

Because capture never starts:

- No purple screen-recording indicator on the Mac
- iPad stays on the OpenDisplay receiver UI (does not show the extended desktop)

This is **not a network issue**: TCP connect + `phone hello` succeed before the failure.

Related existing discussion: #107 (same `MacSender` Code=3 path; that report was more about reconnect races / USB flakiness). This report focuses on a **persistent** Code=3 on a clean session, plus a **working local workaround** (CGDisplayStream fallback).

## Environment

| Item | Value |
|------|--------|
| Mac | Mac mini, Apple M2 |
| macOS | 15.7.x (Sequoia) |
| Displays | One external 4K monitor only (no built-in panel) |
| OpenDisplay Mac | 0.11.0 (release) |
| OpenDisplay iOS | App Store / TestFlight receiver |
| Transport | Wi‑Fi (reproduced; USB not required for this failure) |
| Mode | Extend |

## Symptoms (user-visible)

1. First install / first successful session sometimes **did** show the purple recording indicator and stream to the iPad.
2. Later sessions: iPad may connect, Mac may briefly show a second display arrangement, but **no purple indicator** and iPad remains on the OpenDisplay app chrome.
3. Mac UI status (release): red/failed with  
   `WiFi · Failed: virtual display never appeared in SCShareableContent`.

## Evidence from `/tmp/opensidecar-mac.log` (release)

```text
connection ready to iPad
Connected to iPad
phone hello: 1640x2360 @2.0x
virtual display created: id=… 820x1180pt @2x
sender failed to start: Error Domain=MacSender Code=3
"virtual display never appeared in SCShareableContent"
```

With extra instrumentation (local build), during the `findSCDisplay` poll:

```text
findSCDisplay attempt N: sck=[1] ns=[1] want=<vdId> online=0
```

Interpretation:

- `SCShareableContent` only ever listed the **main** display (`sck=[1]`).
[...]

PR #181, opened 2026-07-30
fix(mac): fall back when ScreenCaptureKit misses a virtual display

jcxl8:
## What changed

- fall back to `CGDisplayStream` when a newly created `CGVirtualDisplay` never appears in `SCShareableContent`
- re-check readiness after one settings re-apply and return to ScreenCaptureKit when the display becomes online
- keep the fallback limited to the typed virtual-display enumeration failure
- reuse the existing frame processing, encoder backpressure, cursor, rotation, and status paths
- isolate IOSurface ownership per display stream and retain the stream until CoreGraphics delivers `.stopped`
- convert `CGDisplayStream` host timestamps with `CMClockMakeHostTimeFromSystemUnits`
- make stream shutdown idempotent and prevent stopped streams from submitting late frames

## Why

On some macOS configurations, the virtual display is created successfully but needs one settings re-apply before WindowServer brings it online. ScreenCaptureKit can fail to enumerate a display that remains offline, while forcing a late-online display through `CGDisplayStream` can produce black frames. The sender can otherwise fail with:

```text
virtual display never appeared in SCShareableContent
```

This leaves the receiver connected but without an extended desktop. The deprecated CoreGraphics stream remains available for displays that are still offline after re-apply; displays that become online use ScreenCaptureKit.

Addresses #142.

## User impact

Users affected by the ScreenCaptureKit enumeration failure or the late-online black screen can extend their desktop to an iPad or iPhone. The normal ScreenCaptureKit path is used whenever the display is online after the readiness retry.

## Validation

- macOS Release build with Xcode succeeded
- strict deep code-signature verification passed
- live Wi-Fi test extended an iPad at `2224×1668`
- pixel probe reproduced fully black `CGDisplayStream` output on a late-online virtual display (`avg=0`)
- live regression confirmed the late-online display selects ScreenCaptureKit and delivers frames
- black-screen route check passed (`GREEN: online virtual display uses ScreenCaptureKit`)
- Screen Recording, Accessibility, and Local Network permissions verified
- fallback lifecycle and timestamp handling reviewed separately for spec and code quality

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -18,10 +18,22 @@ import VideoToolbox
 import Network
 import CoreMedia
 import AppKit
+import IOSurface
 
 enum CaptureMode: String {
     case mirror   // main display (Milestone 1)
     case extend   // virtual display (Milestone 2)
+}
+
+private enum SenderError: LocalizedError {
+    case virtualDisplayNotShareable
+
+    var errorDescription: String? {
+        switch self {
+        case .virtualDisplayNotShareable:
+            return "virtual display never appeared in SCShareableContent"
+        }
+    }
 }
 
 /// Capture-resolution / bitrate trade-off. The virtual display always runs at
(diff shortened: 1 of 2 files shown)


peetzweg:
Is this still the case? I think I might have fixed that a while ago. Reopen if still an issue, thanks!

PR #207, opened 2026-08-08
fix: Prevent Mac & iOS device displays from diverging due to dropped frames

jlfwong:
There's a specific series of event that can happen which can cause the mirror device to display something different than the Mac display does indefinitely.

1. Mac encodes a burst of frames, then skips encoding many of them due to backpressure on encoding or sending
2. iPad receives the the ones before we started skipping

Normally this isn't problem, because as soon as something causes the Mac to encode a new frame, it will send that update to the iPad and things look correct

The problem is if we dropped the *last* frame and then the Mac virtual display stops changing. In this case, nothing causes another encode, and the iPad is permanently stuck (at least until something changes on the mac virtual display).

The fix for it is to ensure that, whenever we drop a frame, that we encode at least one more frame after that dropped frame.

The way I did that in this PR was to use a timer. This works fine. There's probably a more elegant solution which doesn't rely on a timer and notices when backpressure has fallen and then re-transmits the last pixel buffer then. Let me know if you'd rather go with that approach.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -255,6 +255,9 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // keyframe on — so keep the last frame around and re-encode it.
     private var lastPixelBuffer: CVPixelBuffer?
     private var lastCaptureAt = Date.distantPast
+    /// Debounced replay after encoder/send backpressure drops a frame.
+    /// At most one timer is active; each new drop resets the 30ms deadline.
+    private var dropReplayTimer: DispatchSourceTimer?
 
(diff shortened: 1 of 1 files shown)


peetzweg:
Is there a way to make this issue happen on purpose? 🤔

jlfwong:
I can try to make a demo page where it happens another day, but on a guess: make a page which runs a continuous animation changing the color of something between green and blue at 60fps, then on the last frame, change it to red, and then stop the animation.

I haven't tested that specific scenario, but my guess is that, some of the time, the iOS device will end on a green/blue frame instead of a red frame.

Issue #206, opened 2026-08-07
Extend mode fails on macOS Tahoe 26.6.1: "virtual display never appeared in SCShareableContent"


fabiobbbbbb:
## Environment

- OpenDisplay: 1.15.0
- macOS: Tahoe 26.6.1 (stable)
- Mac: MacBook Air M4
- iPad: iPad A16
- iPadOS: 26.6

## Problem

Mirror mode works correctly.

Extend mode consistently fails with the following error:

```
Failed: virtual display never appeared in SCShareableContent
```

## Expected behavior

The iPad should be connected as an extended display.

## What I tested

- Mirror mode works correctly.
- Extend mode always fails.
- Apple Sidecar (Continuity Display) works correctly in Extend mode on the same Mac and iPad.
- Screen Recording permission is granted.
- Accessibility permission is granted.
- AWDL and Bonjour are working correctly.
- No VPN is active while testing.
- Both Mac and iPad have been restarted.
- Running the latest available version of OpenDisplay (1.15.0).

## Logs

I've attached the log captured during a single failed Extend attempt.

From the log, OpenDisplay successfully queries `SCShareableContent`, but only the physical display is returned. The virtual display never appears before the timeout, which results in the error shown above.

[log.txt](https://github.com/user-attachments/files/30845865/log.txt)

BLACKIELF:
Reproduced this exact failure on another M4 Pro Mac running macOS 26.6.1 (25G76), with OpenDisplay 1.16.1: Mirror worked, while Extend created a `CGVirtualDisplay` whose ID never entered `CGGetOnlineDisplayList` or `SCShareableContent`.

A strict single-variable harness kept the descriptor and serial fixed:

```text
productID 0x4F53 -> apply=true, offline/inactive for 10 s
productID 0x4F44 -> apply=true, online/active immediately, visible to SCK
```

Changing only the physical-size/mode profile while retaining `0x4F53` did not help. A rebuilt OpenDisplay using `0x4F44` then extended successfully at 2388x1668 @60 Hz.

I posted the full measurements and upstream handling suggestion on PR #126: https://github.com/peetzweg/opendisplay/pull/126#issuecomment-5281450054

chpuget:
Got exactly the same issue this week following update from MacOS 26.6 to 26.6.1, on a MacBook M5 Pro (Mirror mode works, not Extend mode)

parkamonster:
[...]

PR #217, opened 2026-08-13
fix(macOS): key the virtual display identity to the device, not the transport

lyjj0912:
## Summary

Derive the virtual display serial from the receiver's install id (`PhoneInfo.id`) instead of `ConnectionTarget.sessionID`, so one physical device keeps one macOS display identity across USB and WiFi. Receivers old enough not to send an id keep the historical session-derived serial.

## Why

macOS keys saved display state on vendor/product/serial. OpenDisplay derives that serial from `ConnectionTarget.sessionID`, which *is* the transport — `"usb:<udid>"` or `"wifi:<name>"`. The same iPad therefore presents as two unrelated monitors depending on how it connected, so:

- arrangement and resolution memory are built twice, once per transport
- the identity offset added in #231 is persisted per session id (`displaySerialBump.<sessionID>`), so escaping a poisoned identity over USB teaches the WiFi path nothing
- `upgradeToUSB()` crosses from one identity to the other mid-session

`PhoneInfo.id` is already the transport-independent receiver install id, and `DisplayArrangement` already trusts it as the device key (#116) for exactly this reason. This makes the serial agree with it.

## What this is not

This PR originally also carried its own poisoned-identity recovery. #231 landed a more thorough version of that — offset probing, a matching productID bump, `findSCDisplay` as the online proof, and a dedicated error when every probe fails — so that half is dropped rather than duplicated. The #231 probe loop is untouched here and still owns recovery; it simply probes offsets from a device-stable base instead of a transport-stable one.

## Known limitation

The offset itself is still persisted per session id. With a device-stable base serial, a device that escaped to `+1` over USB starts at `+0` again on its first WiFi session and re-probes, so recovery is one cycle late rather than immediate. Moving that key to the install id means threading `info.id` back to the controller, which only learns it after the hello — deliberately left out of this change.

## Validation

- `xcrun swiftc -typecheck` over `Mac/*.swift Shared/*.swift`: exit 0, 0 errors, 22 warnings — identical counts to untouched `main`, so no new diagnostics (Sparkle stubbed; this machine has Command Line Tools, not Xcode)
- `git diff --check`: clean
- rebased onto `main`; `main` is an ancestor of the branch, so the merge is a fast-forward

Not run here: a real app build and on-device verification, for lack of a full Xcode install on this machine. The earlier hardware validation on macOS 26.6.1 covered the superseded recovery code, not this diff, so I am not carrying those claims over.

### DIFF
--- a/Mac/DisplayIdentity.swift
+++ b/Mac/DisplayIdentity.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+/// Owns the virtual-monitor serial independently of the current transport.
+///
+/// macOS keys saved display state — arrangement, resolution, the hostile
+/// "never come online again" state `MacSender.setupExtend` probes around —
+/// on vendor/product/serial. Deriving that serial from the connection's
+/// session id therefore hands one physical device two unrelated identities,
+/// one for USB and one for WiFi. New receivers report a stable per-install
+/// id in `PhoneInfo.id`, which `DisplayArrangement` already trusts as the
+/// device key; using it here puts both transports on one identity.
+enum DisplayIdentity {
+    /// FNV-1a over the install id. Namespaced so the value cannot collide
+    /// with the session-derived serials older receivers still get.
+    static func serial(for deviceID: String) -> UInt32 {
+        var hash: UInt32 = 2_166_136_261
+        for byte in "device:\(deviceID)".utf8 {
+            hash = (hash ^ UInt32(byte)) &* 16_777_619
+        }
+        return hash == 0 ? 1 : hash
+    }
+}
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -383,10 +383,11 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
         let displayName = endpointName.hasPrefix("iPhone / iPad")
             ? "OpenDisplay — \(info.kind)"
             : "OpenDisplay — \(endpointName)"
-        // Keep one stable identity across rotations. Reconfiguration below
-        // applies a new mode to the existing virtual monitor, so macOS keeps
-        // its windows and arrangement attached to this physical device.
-        let serial = displaySerial
+        // Transport and display identity are separate concerns. New receivers
+        // identify the physical install in `info.id`, so USB/WiFi must use the
+        // same virtual-monitor serial. Older receivers retain the historical
+        // session-derived fallback.
+        let serial = info.id.map { DisplayIdentity.serial(for: $0) } ?? displaySerial
         // Arrangement memory (#116): keyed on the device's install id so the
         // display returns to its spot across transports and orientations —
         // the serial-keyed memory macOS keeps starts from scratch whenever


wantaekchoi:
I ran into this failure independently today on a Mac→Mac setup, so I spent some
time measuring the thing this PR is built on. Good news first: the premise
reproduces cleanly on my hardware too.

Sender: MacBook Pro (Mac17,2, M5), macOS 26.5.2 (25F84), OpenDisplay 1.16.1
release build. Receiver: iMac21,2 running a build of #145, wired.

## The premise reproduces

Same machine, same minute, same `quality`; only the connection path differs.

- over Bonjour (`status[wifi:…]`): 8 consecutive failures, display ids 29-34 and
  40, none ever appearing in `CGGetOnlineDisplayList`
- with the `-host`/`-port` override (`status[usb:first]`): succeeds immediately
  and has stayed up since

```
[16:26:02.465] capture started: 3360x1890 display 41 generation 1 mode extend
[16:26:02.465] status[usb:first]: Extending to Mac (3360×1890)
```

As a control, a standalone binary building the same descriptor
`VirtualDisplay.swift` builds — same 2240x1260pt @2x, same
`maxPixelsWide/maxPixelsHigh`, same `sizeInMillimeters`, same name — registered
on the first poll every time, including while the app was failing. The private
API is healthy on this machine.

The difference the code shows between the two paths is the serial.
`displaySerial(for:)` hashes `target.sessionID`, which is `"usb:first"` for the
override and `"wifi:<service name>"` for Bonjour, so each path derives its own
value and keeps it across reconnects. The transports of course differ in other
ways as well, and I have not shown the serial is the only variable, so I would
not call this proof. But one derived serial failing permanently while another
succeeds on the same machine at the same moment is at least consistent with what
this PR describes.

## `CGDisplayIsOnline(id) != 0` is true for a display that never registered

The other half is on the detection side. `boolean_t` is `Int32`, and
CoreGraphics returns **-1**, not 0, for a display ID it does not know:

```
id                                CGDisplayIsOnline   != 0    in CGGetOnlineDisplayList
--------------------------------  -----------------   -----   ------------------------
1   (built-in, online)                            1   true    true
7, 42, 999, 123456 (never used)                  -1   true    false
46  (real CGVirtualDisplay id,                   -1   true    false
     after the object was released)
```

The last row is the one that matters. That is an id WindowServer itself assigned
to a `CGVirtualDisplay` whose `apply()` returned true; once the display is no
longer online, the call returns -1 for three seconds straight, exactly as it does
for an id that never existed. A display held alive returns 1. So on this build
"not online" comes back as -1, never as a clean 0.

```swift
import CoreGraphics
for id: CGDirectDisplayID in [1, 42, 999] {
    print(id, CGDisplayIsOnline(id), CGDisplayIsOnline(id) != 0)
}
// 1 1 true / 42 -1 true / 999 -1 true
```

So in `waitForDisplayOnline` the first poll returns `true` for exactly the state
this PR targets — a display that never registered with WindowServer, whose id
WindowServer therefore does not recognise. `vd = candidate; break identityLoop`
runs, the rollover never executes, and control reaches the unchanged
`findSCDisplay`, which fails with the same Code=3 as before.

`== 1` fixes the immediate bug. Testing membership in `CGGetOnlineDisplayList` is
stronger, since it cannot be fooled by an error return and it is what the sampler
in #142 actually measured.

The same pattern sits at `VirtualDisplay.swift:224`:
`guard CGDisplayIsInMirrorSet(id) != 0 else { return }` also passes on an
unregistered display, so `ensureNotMirrored()` then runs a
`CGBeginDisplayConfiguration` transaction against a non-existent display roughly
every 200 ms for as long as the retry loop runs.

## Smaller things

- #206 looks like the same report (Tahoe 26.6.1, same error string, filed 7 Aug)
  and is not referenced — worth linking.
- `NSError(domain: "MacSender", code: 4)` collides with the existing code 4 at
  `MacSender.swift:1251` on `main`.
- `Mac/DisplayIdentity.swift` is not in the test target's file list in
  `project.yml`, so a test for it would not compile in.

Happy to run anything else against this hardware — I have the failing state and
a standalone harness for the private API.

wantaekchoi:
I tested this PR's premise directly and it doesn't hold on my machine either — same result @RedVikingSound reported on theirs.

```
serial=FAC17F0B  apply=true  id=49  online=YES   <- the Bonjour-path serial that failed 8x here
serial=804628D5  apply=true  id=50  online=YES   <- the usb:first serial that works
serial=11112222  apply=true  id=51  online=YES
serial=33334444  apply=true  id=52  online=YES
```

`FAC17F0B` is FNV-1a over the exact `sessionID` those failing sessions used, run on the same machine with no reboot since the failures. It registers normally.

My earlier "standalone control" didn't match the serial — I listed the descriptor fields it copied but used an arbitrary serial, so it never tested this. That was my mistake, and it made the transport correlation look like support for the premise when it isn't.

The detection bug is separate and still worth fixing on its own: `CGDisplayIsOnline` returns -1 for an id WindowServer doesn't know, so `waitForDisplayOnline`'s `!= 0` is true for a display that never registered, whatever the cause. Membership in `CGGetOnlineDisplayList` can't be fooled that way.

PR #219, opened 2026-08-15
fix(mac): rebuild stale offline virtual displays

BLACKIELF:
## Summary

- require a virtual display to be online and active before reusing it
- rebuild stale/offline displays instead of re-attaching capture to them
- use a fresh product ID on each rebuild to bypass stale WindowServer display-identity records

## Root cause

A stale `CGVirtualDisplay` can retain non-empty `CGDisplayBounds` while `CGDisplayIsOnline` and `CGDisplayIsActive` are both false. The existing recovery path treated non-empty bounds as proof that the display was live, then attempted to re-attach ScreenCaptureKit to an offline display. Extend mode consequently stayed black because the display never appeared in `SCShareableContent`, while mirror mode continued to work by capturing the built-in display.

This change centralizes the liveness check in `VirtualDisplay.isUsable`. Offline or inactive displays now take the existing full-rebuild path, and a fresh product ID prevents the rebuild from resolving to the same stale WindowServer identity. Display placement remains handled independently by `DisplayArrangement`.

## Validation

- `swiftc -frontend -parse Mac/VirtualDisplay.swift`
- `swiftc -frontend -parse Mac/MacSender.swift`
- `git diff --check`
- live test on Apple Silicon / macOS 26.6: Mirror → Extend created a new virtual display ID
- `system_profiler` reported the rebuilt display as `Online: Yes`, `Mirror: Off`, `2388 x 1668`, `60 Hz`
- USB client stream stabilized at 59–60 FPS with no black screen

`CGVirtualDisplay` remains a private macOS API, so future system releases may still require compatibility adjustments.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -466,7 +466,7 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     /// Returns false when there is no reusable display or macOS rejected the
     /// mode switch, letting the caller use the legacy rebuild fallback.
     private func resizeExistingDisplay(for info: PhoneInfo) async throws -> Bool {
-        guard let vd = virtualDisplay else { return false }
+        guard let vd = virtualDisplay, vd.isUsable else { return false }
 
         let pointsWide = (info.pixelsWide / 2) & ~1
         let pointsHigh = (info.pixelsHigh / 2) & ~1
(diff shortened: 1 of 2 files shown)


peetzweg:
PR #288 handles poisoned identities during initial display setup, but it does not make this PR redundant: the capture-recovery path still uses non-empty `CGDisplayBounds` as its first reusable-display check. The online/active liveness test remains useful. Please rebase and retain that check while adapting the rebuild identity behavior to #288’s persisted serial/product-offset recovery.

Issue #224, opened 2026-08-18
Proposal: Android support via USB Accessory (AOA)


jpbhdrey:
Today wired connect only works for iOS/iPadOS, through the built-in `usbmuxd`. Android has no equivalent and the only way in right now is `adb forward` plus the `-host`/`-port` escape hatch. So developer options, USB debugging, an RSA prompt, and a terminal command. 

AOA (Android Open Accessory) gets to the same "plug and play" the iOS app already has, allowing the user to seamless spin up a new screen. After pluggin Android shows a plain "Open OpenDisplay to handle this USB accessory?" prompt without needing to add developer options in the config.

I know that this repo doesn't support Android but I have a working fork at PR #225 tested with a Galaxy Tab S9 FE+ (see screenshots at josepacelli/opendisplay-android#32). It would couple two different repos, but maybe this can be thought as part of the general protocol discussed in #15.

peetzweg:
Reasonable ask. Will try to look into this next week.

jpbhdrey:
Just an addition. The linked PR is misleading when I said that the Mac client code would need to hardcode the android client's repo or package id. Given that, it would be an agnostic addition that every android implementation could follow. I will update that pr shortly

PR #225, opened 2026-08-18
feat: connect Android over USB without adb

jpbhdrey:
This PR does two things: abstracts the transport behind a `ByteStream` protocol (WiFi/usbmux vs. USB accessory), then adds a USB accessory (AOA) transport for Android. Related to #224 .

No UI yet, so you need to enable with `defaults write com.peetzweg.opensidecar.mac.debug accessory -bool YES`.

I verified it in a Galaxy Tab S9 FE+.

The AOA identity strings (`Mac/USBAccessory.swift`) must match exactly the Android app's `accessory_filter.xml`, tying this PR to one specific implementation (josepacelli/opendisplay-android#33) which is not an official/first-party one. Maybe this should be discussed together with the Android client repo maintainer @josepacelli.

Also, I didn't implement a watcher for a new Android device connected after the app is running. The handshake is tried when OpenDisplay starts. At most we can reconnect to a previously connected Android AOA in the same section. We could iterate in every new usb devices connected from some MacOS notification, so it could be done.

### DIFF
--- a/Mac/ByteStream.swift
+++ b/Mac/ByteStream.swift
@@ -1,0 +1,43 @@
+// Transport abstraction: send bytes, read exactly N bytes, close.
+// Wire framing is [4-byte BE length][payload].
+
+import Foundation
+import Network
+
+protocol ByteStream: AnyObject {
+    /// `completion` fires once `data` is handed off to the transport.
+    func send(_ data: Data, completion: @escaping (Error?) -> Void)
+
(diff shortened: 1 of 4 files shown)


jpbhdrey:
There are some problems that I found afterwards, so it isn't a complete implementation (for instance, the watcher for new devices isn't implemented yet). If feedback is given and you decide to add this feature, I will keep working on it

Issue #125, opened 2026-07-10
Wrong local network access status


christophby:
The Mac app displays a warning that it does not have access to the local network, even though access is actually available.

<img width="1088" height="678" alt="Image" src="https://github.com/user-attachments/assets/f6eccabb-f133-46bb-82b5-4a50b1806344" />

CaioFaSoares:
Currently having the same issue. Tried with both 1.12.0 and 1.11.0 

Happened after updating but rolling back did not solve it.

peetzweg:
But it's just cosmetic right? The last release was more of a plumbing yet again. Next releases will be quicker. Putting this up next for sure.

christophby:
>But it's just cosmetic right? 

yep

youngledo:
Any progress?

peetzweg:
Will have some time this week to make a bunch of improvements. Will put this on the short list to finally fix this.

Issue #120, opened 2026-07-08
Unable to find iPad over Wifi


peetzweg:
https://www.reddit.com/r/ProductivityApps/comments/1un1342/comment/owc4iaj/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

> Hi man, I'm currently using it right now. It's been a big help and I'm so hyped by this discovery! I'm over at my boyfriend's place a lot and thought to use his Ipad as my second monitor for work -but because of the different apple ID, it requires me to download a thirdparty app that I'm not willing to splurge a shit ton of money on, so your app definitely was a huge huge huge help! There were a few obstacles I encounter -like not being able to quickly load up the screen or make it to function using the wifi feature (I'm using a USB cord to connect my mac to his ipad) but overall it's been great! big ups to you man, can't wait to see what's more in store : )

> My macbook and my bf's Ipad are connected to the same wifi but for some reason -even if his ipad name show up on the wifi names on my mac, it still won't allow me to connect. that's pretty much it, i've tried renaming it, clicking on extending or mirroring --nothing. so yeah, would love for that to get fix ! :)

peetzweg:
Worth ruling out client isolation on that WiFi network first.

Some networks (guest WiFi, hotels, most conference and office networks, and home routers with guest mode left on) forward each client's traffic upstream and drop anything addressed sideways to another client on the same subnet. The internet keeps working perfectly, so nothing looks wrong. Many access points still flood mDNS while blocking unicast, which produces exactly what you describe: the iPad's name shows up in the Mac app, and every connect attempt goes nowhere. Renaming, extend vs mirror, none of it can help, because the packets never leave the access point.

A quick way to tell, on the Mac:

```sh
ipconfig getoption en0 router     # prints your router's address
ping -c 3 <that address>
ping -c 3 1.1.1.1
```

If 1.1.1.1 answers and your own router stays silent, peer to peer traffic is being dropped. No permission grant or app update fixes that one, and USB is the way in (it is lower latency anyway, so nothing is lost). A personal hotspot from the phone also works, since you own that network.

The next Mac release detects this and says so in the Devices list instead of blaming Local Network permission. It also writes the verdict to `~/Library/Logs/OpenDisplay/opendisplay.log` at launch:

```
network: blocked — gateway silent, internet up
```

If you are still using it, that line would confirm whether this was the cause.

PR #226, opened 2026-08-19
feat(mac): detect networks that block device-to-device traffic

peetzweg:
Relates to https://github.com/peetzweg/opendisplay/issues/125 and https://github.com/peetzweg/opendisplay/issues/120

The Mac app now probes whether it can reach anything on its own subnet, and says so in the Devices list instead of blaming Local Network permission.

**Why:** guest, hotel and corporate WiFi commonly runs client isolation. Each client reaches the internet and nothing else, so Bonjour advertisements never arrive and a direct dial to the device can never land. Nothing on either device fixes it, but the app was pointing users at a Local Network settings pane that could not help. Verified live on a network doing exactly this: the internet answers in 5ms while the router and all six LAN neighbours drop every packet on every port.

**How:** `NetworkDiagnostics` pings the router, falls back to a TCP dial (a refusal proves reach as well as a connect does), and compares against a public resolver. Internet up while our own router is mute is the signature. The judgment call worth reviewing is how conservative to be: a probe macOS refuses to send reports `inconclusive` rather than `silent`, so a denied Local Network grant cannot masquerade as a hostile AP, and a full-tunnel VPN owning the default route is not judged at all. That costs some detection on hardened-but-fine networks, which felt like the right trade against replacing one wrong diagnosis with another.

The verdict is logged at launch, so attached bug reports answer "was this one of those networks?" without a reproduction:

```
network: blocked — gateway silent, internet up
```

README gains a troubleshooting section covering the symptoms, a two-command self-check, and when the existing `defaults write host` escape hatch does and does not help. 19 tests over the pure logic: the verdict table, the ICMP framing, and the subnet check that keeps Tailscale from being mistaken for the LAN.

### DIFF
--- a/Mac/NetworkDiagnostics.swift
+++ b/Mac/NetworkDiagnostics.swift
@@ -1,0 +1,482 @@
+import Foundation
+import Network
+import SystemConfiguration
+
+/// Why WiFi mode can't see a device.
+///
+/// These need telling apart because the remedies have nothing in common, and
+/// until now every empty device list was blamed on missing Local Network
+/// permission (issue #125) — including on networks where no permission grant
+/// could ever have helped.
+///
+/// The case worth detecting is `.blocked`. Guest WiFi, hotel and conference
+/// APs, and many corporate networks run *client isolation* (a.k.a. AP
+/// isolation): each client's traffic is forwarded upstream, and everything
+/// sideways to another client on the same subnet is dropped. The internet keeps
+/// working perfectly, so nothing looks wrong, while Bonjour advertisements never
+/// arrive and a direct dial to the device can never land. Apple's own Continuity
+/// features (Handoff, Universal Clipboard) break on the same networks, which is
+/// usually the tell a user has already noticed without connecting it to this.
+enum LocalNetworkVerdict: Equatable {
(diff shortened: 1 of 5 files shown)


peetzweg:
### Before merging: verify on a normal network

This was built and tested on a network that *is* locked down, so the `.ok` path has never run outside unit tests. The router here has not answered a single packet, which means the "everything is fine, stay quiet" branch is completely unexercised against reality.

On home WiFi, launch the app and check:

- [ ] No warning row appears in the Devices section.
- [ ] The log says `network: ok — gateway reachable, internet up` (not `blocked`).
- [ ] The **Local Network** row behaves as before, with the original help text rather than the "can't be determined" variant.
- [ ] The phone still appears and connects over WiFi as usual, i.e. the probe traffic has not disturbed anything.

If it reports `blocked` at home, that is a false positive and this should not ship as is. The gateway probe (ping, then a TCP dial on 80/443/53) is a proxy for client isolation rather than a direct test of it, so a router that drops ICMP and closes those ports would trip it. That is the known weak point and this is the check for it.

Worth trying one hostile network too, if you pass through one: a cafe or hotel WiFi should read `blocked`.

Issue #221, opened 2026-08-15
USB connection broken on iPad OS 27 Public Beta 3. "Virtual display never appeared in SCShareableContent"


parkamonster:
I've just updated to M4 iPad Air to iPad OS 27 Public Beta 3 and now I'm seeing connection issues.

Attempting a USB connection now results in an error "USB • Failed: virtual display never appeared in SCShareableContent"

All was fine with previous iPad OS beta versions.

OpenDisplay works fine on my iPhone 17 Pro over USB and WiFi running iOS 27 Public Beta 3 (go figure!). 

My Mac is running the latest Tahoe 26.6.1.

I tried rebooting the Mac and iPad but to no avail.

The only solution I found was connect to OpenDisplay over WiFi (which still works) then plug the USB cable in. I've attached my log.

[opendisplay.log.zip](https://github.com/user-attachments/files/31105691/opendisplay.log.zip)

peetzweg:
Thanks for the report. Thanks for the logs. I havne't yet looked into supporting the next version of iOS and Mac OS. September release is coming closer for sure. This will be on my plate pretty soon. Fingers crossed Apple is not crippling APIs used by OpenDisplay.

peetzweg:
Merged PR #288 adds bounded recovery for this class of virtual-display failure: it starts large displays on a conservative canvas, verifies that ScreenCaptureKit sees the display, promotes the same identity in place, and tries a fresh persisted identity when macOS has poisoned the saved one. That is directly relevant to `virtual display never appeared in SCShareableContent`, but this exact iPadOS 27/Tahoe USB-first sequence was not reproduced. Please retry a build from current `main` (or 1.20 once released) and attach a fresh log if USB still requires connecting over Wi-Fi first. Keeping this open pending that confirmation.

PR #231, opened 2026-08-20
fix: Honor the system capture stop and survive poisoned display identities

### LINKED ISSUE #206: Extend mode fails on macOS Tahoe 26.6.1: "virtual display never appeared in SCShareableContent"

### LINKED ISSUE #230: "Stop Extending" from the system UI permanently breaks reconnecting: WindowServer saves a poisoned config for the virtual display identity
## Symptom

After stopping a session from the macOS system UI (the purple screen indicator in the menu bar / Control Center, "Stop Extending") instead of the in-app Disconnect, the device can never connect again. Every attempt ends with:

```
Failed: virtual display never appeared in SCShareableContent
```

with intermittent `CGVirtualDisplay applySettings FAILED` lines in the log. The failure survives:

- restarting the Mac app and the receiver app
- `tccutil reset ScreenCapture` plus re-granting Screen Recording
- restarting `replayd`
- a display sleep/wake cycle

This is most likely what #206 and #221 are hitting (same error string, same persistence across restarts).

## Root cause

Reproduced and diagnosed live on macOS 26 (Darwin 25.6.0), OpenDisplay 1.16.1:

1. The user presses stop in the system UI. The SCStream stops with `SCStreamErrorDomain -3815` ("Failed to find any displays or windows to capture").
2. The capture recovery loop (`scheduleCaptureRecovery`, 3s interval, unbounded) does not classify the error and re-attaches capture ~20s later, effectively defying the user's stop.
3. WindowServer saves a display configuration for the virtual display's identity with `IndependentOutput: true` (and a Scale 1 / Depth 4 mode) in `~/Library/Preferences/ByHost/com.apple.windowserver.displays.<host-uuid>.plist` under `DisplaySets.Configs`.
4. Our virtual display identity is deterministic (vendor `0x5043`, product `0x4F53`, serial = FNV-1a of the session id in `SenderController.displaySerial`), so the display UUID is stable, and WindowServer re-applies the poisoned config on every recreate: the display object is created, but it never joins the active display list and never shows up in `SCShareableContent`.

Proof by differential test: a WiFi session to the same phone (different session id, therefore a different serial, display UUID, and saved config) extended fine immediately, while the USB identity kept failing, in the same app process with the same TCC grant.

Removing the poisoned entry from the plist is not enough on its own: WindowServer holds the config in memory and only re-reads it at login, so manual recovery needs the plist edit plus a logout or reboot. That is not something we can ask users to do, hence the app-side fix.

## Fix (PR follows)

- `setupExtend` abandons a display identity that never comes online and retries with serial+1, persisting the working offset per device. Verified on the still-wedged machine: serial+0 times out, serial+1 comes online in under a second, extend works; the next launch uses the persisted offset directly.
- "SCShareableContent lists zero displays" (capture permission broken app-wide) is now distinguished from "our display is missing" (poisoned identity), so fallback identities are not burned on permission problems and the error message tells the user what to actually do.
- `SCStreamError.userStopped` is treated as a disconnect (including the auto-connect opt-out), not as a fault to retry. The defiant auto-restart is the likely trigger for the poisoning.
- The capture recovery loop is bounded (5 rounds), instead of hammering WindowServer with create/destroy cycles forever.
- A session whose `sender.start()` threw used to stay parked with a leaked virtual display (still holding the device's serial) and swallowed every later connect for that device until a manual Disconnect. It now frees its pipeline and can be replaced by a fresh connect or restarted from its row.

peetzweg:
Fixes #230. Most likely also resolves #206 and #221 (same error, same persistence).

Stopping a session from the macOS system UI ("Stop Extending") makes WindowServer save a hostile config (`IndependentOutput`) for our deterministic display serial, and it re-applies that config on every recreate: the display never comes online again and every session fails with "virtual display never appeared in SCShareableContent". This PR makes the Mac app survive that state and stop provoking it.

**Why:** users who stop via the system indicator instead of the in-app Disconnect end up permanently unable to reconnect, across app restarts and even permission resets. Full diagnosis in #230.

**How:** `setupExtend` abandons an identity that never comes online and retries with serial+1, persisting the offset per device. `SCStreamError.userStopped` is now honored as a disconnect (with auto-connect opt-out) instead of auto-restarting capture against the user's intent, the capture recovery loop is bounded at 5 rounds, and a session whose `start()` threw frees its pipeline instead of leaking the display and swallowing every later connect. Judgment call: an identity bump is permanent per device; macOS-side arrangement starts fresh for the new serial, but our install-id keyed arrangement memory (#116) restores placement.

Verified on the poisoned machine: serial+0 times out, serial+1 extends in under a second, next launch extends in 0.6s via the persisted offset. Existing test suite passes (34).

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -111,9 +111,12 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // the session — teardown plus auto-connect opt-out — so the app honors
     // the stop instead of fighting it.
     @MainActor var onCaptureStoppedByUser: (() -> Void)?
-    // Fired when the device's base display identity had to be abandoned
-    // (macOS saved hostile state for it — see setupExtend) and a bumped
-    // serial worked: carries the offset for the controller to persist.
+    // Fired when the device's display identity had to be abandoned (macOS
+    // saved hostile state for it — see setupExtend) and a bumped identity
+    // came online instead: carries the validated TOTAL offset from the
+    // device's base identity, for the controller to store as-is. Absolute,
+    // not a delta — repeated bumps in one session must not accumulate into
+    // an offset nothing ever validated.
     @MainActor var onDisplayIdentityBumped: ((UInt32) -> Void)?
 
     private var stream: SCStream?
@@ -132,6 +135,12 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
     // Stable per-device serial for the virtual display, so macOS can tell
     // multiple OpenDisplay monitors apart and persist their arrangement.
     private let displaySerial: UInt32
+    // How far this device's identity has already moved off its base serial
+    // and productID (identities macOS saved hostile state for are abandoned
+    // permanently — see setupExtend). Advanced in-session when a fallback
+    // identity is validated, so a rotation rebuild doesn't re-probe the
+    // poisoned one.
+    private var baseIdentityOffset: UInt32
 
     // ── Encoder parallelism limiter (maxPendingEncodes = 1) ─────────────────
     //
@@ -279,12 +288,13 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
 
     init(transport: SenderTransport, name: String, mode: CaptureMode,
          quality: StreamQuality = .best, displaySerial: UInt32 = 0x0001,
-         awaitingWake: Bool = false) {
+         identityOffset: UInt32 = 0, awaitingWake: Bool = false) {
         self.transport = transport
         self.endpointName = name
         self.mode = mode
         self.quality = quality
         self.displaySerial = displaySerial
+        self.baseIdentityOffset = identityOffset
         self.awaitingWake = awaitingWake
         super.init()
     }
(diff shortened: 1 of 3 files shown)


peetzweg:
Live verification of the userStopped path on the machine that originally hit the bug: pressing "Stop Extending" in the system UI now delivers SCStreamError -3817, the app honors it as a disconnect (clean teardown, auto-connect opt-out, no capture restart, no poisoning), and a manual Connect afterwards re-extends in under a second on a clean identity.

```
[11:09:16.372] stream stopped with error: ... Code=-3817 "The user stopped the stream"
[11:09:16.372] session usb:... capture stopped via the system UI — honoring as disconnect
[11:11:40.684] phone hello: 1179x2556 @3.0x
[11:11:41.150] status[usb:...]: Extending to iPhone (1176×2556)
```

Note: yesterday the same button press produced -3815 cascades instead of -3817. That was most likely already fallout from the old defiant capture restart after an earlier stop, which is exactly the loop this PR removes.

peetzweg:
Addressed the review findings in d236411, each verified live on the affected machine where possible:

- S1: userStopped is only honored from an interactive console. Live lock test: locking delivers -3815 here (not -3817), recovery ran with the new lock-aware error text, and the extension came back by itself ~20s later on unlock. No regression vs main.
- S2: recovery-failure counter resets whenever capture comes back up.
- S3: the identity offset is now persisted as a validated absolute value (no accumulation across rotation rebuilds), and only on created-but-never-surfaced evidence, never when creation is refused by a twin still holding the serial.
- S4: failed sessions are excluded from cross-transport covering and from dedupe, so a corpse can neither block auto-connect on the other transport nor win "keep the cable" against a working WiFi session.
- P1: the productID moves together with the serial (0x4F53 + offset), covering the product-keyed poisoning reported in #206. Re-verified the fallback against the poisoned machine with the double bump: base identity times out, serial+1/product 0x4F54 comes online in under a second.
- P2: exhausting all fallback identities now reports "saved display state in macOS is blocking OpenDisplay's displays: log out and back in (or restart the Mac), then reconnect" instead of the bare SCShareableContent message.
- P4: the zero-displays error names the locked screen as the likely benign cause before suggesting a permission re-grant (validated during the lock test, where it fired for exactly that reason).

P3 (whether #221 is the same bug) stays open as a follow-up; the PR wording already hedges.

PR #232, opened 2026-08-20
fix: Cut the silent wait while probing a poisoned display identity

peetzweg:
Follow-up to https://github.com/peetzweg/opendisplay/pull/231, from a field log of the v1.17.0 rollout: the recovery worked first try, but the user sat on a black screen for ~9 seconds with no explanation and mashed reconnect five times.

**Why:** the base identity holds the full ~6.5s surfacing window before the fallback is tried, and the only status that explains what is happening is emitted *after* the fallback already succeeded.

**How:** `findSCDisplay` gains a `polls` parameter. The base identity probe uses a short window (~3-4s); fallback probes keep the patient one, so a Mac that is merely slow still comes online before the identity walk gives up, and a false timeout on the base only moves the serial (arrangement memory keys on the install id, not the serial). Independently, any surfacing wait that passes the one-second mark now reports "Bringing the display online…" instead of leaving the previous status on screen.

34 unit tests pass. The iPad-side half of the choppy experience (silent black screen while connected without video) is a separate iOS PR.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -455,7 +455,15 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
             guard let candidate = created else { continue }
             virtualDisplay = candidate
             do {
-                display = try await findSCDisplay(id: candidate.displayID)
+                // A healthy display surfaces in well under a second; a poisoned
+                // one never does. Wait on a short clock for the first identity
+                // so the user isn't staring at a black device for the full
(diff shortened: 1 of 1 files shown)


peetzweg:
The recovery mechanism changed substantially in merged PR #288, but this UX improvement is not included: current `findSCDisplay` still uses the fixed surfacing window and does not publish the early “Bringing the display online…” progress status while it waits. Please rebase or port the shorter base probe and early status onto current `main`.

PR #238, opened 2026-08-23
feat: iOS setting to fit the display into the safe area (avoid notch & rounded corners)

coreprocess:
> **Before you review:** this is a feature proposal I'd like to gauge interest on
> first. The implementation is complete, but it hasn't been tested on-device yet —
> if this is something you'd want in OpenDisplay, I'll put in the time to test it
> properly (see the testing notes at the bottom) and iterate on review feedback.
> If it doesn't fit your plans for the project, no hard feelings — just say so and
> I'll close the PR.

## What

Adds a "Display area" section to the iOS settings sheet:

- **Avoid notch & rounded corners** toggle (default off — the current edge-to-edge
  behavior is completely unchanged unless you opt in).
- **Extra margin** slider (0–24 pt, default 8 pt), shown while the toggle is on.
  The system safe-area insets don't cover the rounded corners on the long screen
  edges, so this trims a little more off every edge. The value commits debounced
  (with an editing-end accelerator and a removal backstop), because every committed
  value reconfigures the Mac-side display.

With the toggle on, macOS windows, title bars, and the menu bar never hide under
the notch / Dynamic Island or get clipped by the rounded glass corners.

## How it works

The core idea: shrink the **announced panel** instead of letterboxing on the device.

- With the setting on, the phone announces the safe-area sub-rect
  (`window.bounds − safeAreaInsets − 2×margin`, in physical pixels, floored to
  multiples of 4 so the Mac's even-point quantization round-trips exactly) in
  `hello` instead of the full panel. The Mac builds a virtual display that exactly
  matches the visible rect — pixel-perfect fill, no letterboxing, no non-integral
  scaling.
- The video is laid out inside the safe area plus margin. Since the view bounds
  match the announced aspect ratio, aspect-fit fills the rect exactly, and the
  bounds-relative touch/cursor/scroll mapping needed no changes at all.
- The hosting window is reported by the view hierarchy itself (a tiny
  `WindowReader`) rather than scanning scene windows, which can pick the system
  keyboard window. The announcement re-fires on window attach, rotation, setting
  changes, and scene activation; `setPanelPixels` is receiver-queue-confined and
  dedupes by dimensions.
- The status bar is hidden whenever the setting is on (not just while streaming),
  so status-bar-driven insets (iPads, home-button iPhones) don't change between
  the pre-stream `hello` and the streaming layout — avoiding a rebuild at every
  session start on those devices.

## Protocol

One **additive** change, no `pv` bump (per the additive-fields rule in
PROTOCOL.md): new optional `hello` fields `panelWide`/`panelHigh` carry the full
physical panel (long edge first). The Mac reserves the virtual display
descriptor's resize headroom from them — a sub-rect's long edge is
orientation-dependent, so reserving from the first announcement alone would force
a destroy-and-recreate (and the window scatter that comes with it) on the first
rotation or when the setting is turned off. Compatibility both ways: old
receivers omit the fields and get today's behavior; old senders ignore them.
PROTOCOL.md documents the sub-region semantics and the new fields.

## Mac-side hardening

The sender was built on invariants this feature breaks (announced size == physical
panel; orientation-invariant long edge; hellos only on connect/rotation), so a few
latent assumptions needed fixing to make the feature safe:

- The session start path now reconciles the latest `hello` after the initial
  `setupExtend` — a `hello` arriving mid-setup previously updated state but
  triggered nothing (the rebuild path is gated on a live stream).
- All rebuild decisions compare the derived display mode
  (`PhoneInfo.pointsWide/High`, the @2x even-floored points) instead of raw
  pixels, so pixel-only refinements never rebuild a display they wouldn't change.
- `reconfigure` early-outs when the display is already built and capturing at the
  target mode (capture recovery, which reuses it at unchanged dimensions, is
  exempt), reads `lastHello` only through a queue-confined snapshot (all other
  readers now do too), and reports failure: a failed startup reconcile throws like
  a failed initial setup, and a failed debounced rebuild hands off to the bounded
  capture-recovery loop instead of stranding a live-looking session with a
  torn-down pipeline.
- Capture recovery's re-attach branch now requires the surviving display's mode to
  match the latest `hello`; a mismatch falls through to an in-place resize (a
  `hello` landing while capture is down used to strand the session at the stale
  size for good, since the receiver dedupes announcements).
- `VirtualDisplay`'s HiDPI mode enforcement matches modes by width **and** height:
  a portrait sub-rect derives the same point width as the full panel, so
  width-only matching could bless a stale restored full-height mode.

## Notes & testing

- While streaming, the reclaimed notch strip shows black glass (status bar and
  home indicator stay hidden). With the setting on, the status bar is also hidden
  on the idle screen (see above for why).
- On iPads (no notch) the toggle degrades to roughly the margin plus the
  home-indicator strip.
- The margin slider deliberately avoids committing per step: stepped SwiftUI
  sliders don't deliver the closing `onEditingChanged(false)` on iOS 26, and
  VoiceOver adjustments bypass the drag lifecycle, so commits go through a
  debounced `onChange` with backstops.
- CI compiles the Mac target (all `MacSender`/`VirtualDisplay` changes). The iOS
  target isn't built by CI; manual testing should cover: toggling on/off
  mid-stream, rotation in both modes and both settings, dragging the margin
  slider (including via VoiceOver), and verifying touch/cursor alignment and that
  the macOS menu bar is fully visible with the setting on.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -64,9 +64,12 @@ enum StreamQuality: String, CaseIterable {
 }
 
 struct PhoneInfo: Decodable {
-    let pixelsWide: Int   // landscape-oriented (long edge)
-    let pixelsHigh: Int
+    let pixelsWide: Int   // announced display rect (panel or a safe-area
+    let pixelsHigh: Int   // sub-rect), in the panel's current orientation
     let scale: Double
+    let panelWide: Int?   // full physical panel, long edge first — resize
+    let panelHigh: Int?   // headroom (older receivers omit it: the announced
+                          // rect IS the panel there, so it reserves enough)
     let device: String?   // "iPad" / "iPhone" (older receivers omit it)
     let id: String?       // per-install identity (older receivers omit it) —
                           // lets the controller match the same physical device
@@ -76,6 +79,14 @@ struct PhoneInfo: Decodable {
 
     var kind: String { device ?? "device" }
     var protocolVersion: Int { pv ?? WireProtocol.assumedWhenAbsent }
+
+    // The derived virtual-display mode (points at @2x, floored to even for
+    // the encoder). Rebuild decisions MUST compare these, not raw pixels:
+    // hellos can differ in pixels yet agree here — the receiver's init seed
+    // (raw nativeBounds) vs its refined announcement (floored to a multiple
+    // of 4) — and rebuilding for those changes nothing visible.
+    var pointsWide: Int { (pixelsWide / 2) & ~1 }
+    var pointsHigh: Int { (pixelsHigh / 2) & ~1 }
 }
 
 /// How the sender reaches the receiver. Reconnects re-dial from scratch, so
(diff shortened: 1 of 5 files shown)


peetzweg:
Will check it out. I see the value here now. Will make it in soon for sure. 👍

Issue #191, opened 2026-08-02
Pencil strokes have no predicted touch, so they lag behind fingers


peetzweg:
Follow-up from #163, raised by @jlfwong in review.

The finger path forwards UIKit's predicted touch on top of the coalesced samples, which hides roughly one frame of perceived latency:

```swift
if let predicted = event.predictedTouches(for: touch)?.last,
   let n = normalized(predicted.location(in: self)) {
    receiver?.sendTouch(phase: "moved", x: n.x, y: n.y)
}
```

`emitPen` does not. Drawing is exactly where that frame is most noticeable, so the Pencil path is the one that would benefit most and is currently the one going without.

Worth checking whether the prediction should be marked on the wire so the Mac can treat it as provisional, rather than injecting a predicted point as a real tablet sample and leaving a wrong point in the stroke when the prediction misses.

DrNeroCF:
Was wondering about this.  I'll check it, pencil input needs to be more accurate than touch.  On my old iPad Pro, at least, all drawing programs that I've tried besides Notes and Charcoal have latency, PencilKit is definitely hiding some of it, problem is you can't update with the real values later (unless there's something in the Mac-side input).

Astropad attempts to deal with latency by drawing a temporary 'fake' line client-side, but only lets you use pink in the standard edition.

The pure PencilKit drawing just feels so much better than everything else (even in Apple's own Freeform), I think I could put up with a bit of inaccuracy, especially since most apps able to smooth out the input on their end.

Issue #110, opened 2026-07-07
Mirror mode is second-class vs Extend: blurry (1× capture) + touch doesn't work


peetzweg:
## Summary

Mirror mode is a second-class citizen next to Extend — it was built as a "milestone 1" minimal path and never caught up. Two concrete gaps, both confirmed in the code:

1. **Blurry** — it captures the source display at 1× (point) resolution, not native Retina pixels.
2. **No touch/scroll input** — the input injector is only created in Extend mode.

Extend mode has neither problem (pixel-perfect, touch works), so both are straightforward parity fixes.

## 1. Blurry capture — captures at 1× instead of native pixels

`Mac/MacSender.swift:218-220` (mirror path):

```swift
// SCDisplay reports points; capture at point resolution for M1.
let captureW = (Int(Double(display.width) * quality.scale)) & ~1
let captureH = (Int(Double(display.height) * quality.scale)) & ~1
```

`SCDisplay.width/height` are **points**, so on a Retina Mac (e.g. built-in 1512×982 pt / 3024×1964 px) mirror captures at **1512×982** — confirmed by the log: `capture started: 1512x982 ... mode mirror`. That half-resolution frame is then scaled **up** onto the device panel (e.g. iPad 2360×1640) → blur. (Extend captures the virtual display at native pixels, which is why it's sharp.)

Note the aspect-ratio mismatch (16:10 Mac → 4:3 iPad) only produces **letterbox bars**; it is *not* the cause of the blur. The 1× capture is.

**Fix direction:** capture the source display at its native pixel resolution — e.g. `CGDisplayCopyDisplayMode(display.displayID)?.pixelWidth/pixelHeight`, or the display's `NSScreen.backingScaleFactor` — then apply `quality.scale` to that. Implementer's call on whether "Best" should cap at the device panel resolution rather than always the full Mac native (bandwidth), but at minimum it should match the device panel, not 1×.

## 2. Touch/scroll input doesn't work in Mirror mode

`inputInjector` is assigned **only** in `setupExtend` (`Mac/MacSender.swift:281`):

```swift
inputInjector = InputInjector(displayID: vd.displayID)
```

The `.mirror` path (`:211-221`) never creates one, so it stays `nil` and the control handlers are silent no-ops:

```swift
inputInjector?.handleTouch(phase: phase, x: x, y: y)   // :765 — nil in mirror mode
inputInjector?.handleScroll(dx: dx, dy: dy)            // :776 — nil in mirror mode
```

[...]

PR #244, opened 2026-08-27
feat(ios): forward predicted pencil touches to eliminate stroke latency (#191)

kdbhalala:
## Summary

Rescoped per maintainer feedback ([comment](https://github.com/peetzweg/opendisplay/pull/244#issuecomment-5499403668)):
The mirror capture half was independently addressed in #258 (v1.19.0). This PR is now focused strictly on Apple Pencil predicted touches (#191), rebased onto current `main`.

### Apple Pencil Predicted Touches Forwarding (#191)
- In `InputCaptureEngine.emitPen` (`iOS/OpenSidecarPhoneApp.swift`), forwards `event.predictedTouches(for: touch)` during active strokes.
- Uses iOS predictive stroke modeling to eliminate apparent stylus latency during sketching and drawing across network and USB pipelines.

## Verification
- `xcodegen generate && xcodebuild test -project OpenSidecar.xcodeproj -scheme OpenSidecarMac -destination 'platform=macOS'` passes cleanly.
- Rebased cleanly onto latest `main`.

### DIFF
--- a/iOS/OpenSidecarPhoneApp.swift
+++ b/iOS/OpenSidecarPhoneApp.swift
@@ -1041,6 +1041,15 @@ final class InputCaptureEngine: NSObject {
                            azimuth: Double(c.azimuthAngle(in: view)),
                            altitude: Double(c.altitudeAngle))
             }
+            if let predicted = event?.predictedTouches(for: touch) {
+                for p in predicted {
+                    guard let pn = normalize?(p.location(in: view)) else { continue }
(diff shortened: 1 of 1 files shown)


peetzweg:
Half of this PR just shipped independently: #258 (v1.19.0) reads the mirror capture size from CGDisplayMode.pixelWidth/Height, which covers the blur half of #110 that this PR also fixes. Suggest rescoping this to the Apple Pencil predicted-touch work for #191, rebased on current main; the mirror capture hunk will conflict away.

Issue #10, opened 2026-06-10
HEVC encoding option


peetzweg:
## Goal
Better quality per bit than H.264 — most useful in WiFi mode where bandwidth is the constraint.

## Approach
- VideoToolbox HEVC encode on the Mac (hardware-accelerated on all supported Macs), HEVC parameter-set handling (VPS/SPS/PPS) in the Annex B path
- Decode side is the same `AVSampleBufferDisplayLayer` with an HEVC format description
- Negotiate codec in the hello message; fall back to H.264

## Acceptance
Same visual quality at meaningfully lower bitrate over WiFi, no latency regression over USB.

peetzweg:
Decision on how HEVC lands, now that the Mac receiver (v1.18.0) makes big panels a real use case.

Field measurements from #253 changed the picture: H.264 hardware decode stops below 5120 pixels wide on every Mac tested, including an M4 Pro. That is a format ceiling, not a machine-age ceiling, so 5K/6K receivers are impossible on H.264 no matter the hardware. HEVC decodes 5K fine even on a 2017 iMac (measured end to end at ~54 ms over WiFi). That makes HEVC the enabler for exactly the panels the Mac receiver brought into range.

Codec selection policy we settled on:

- H.264 stays the default. It is cheaper to encode, universally decodable, and perfect at phone/tablet stream sizes. No global HEVC switch, no knob.
- The sender switches to HEVC only when all three hold: the stream the receiver asked for exceeds the H.264 decode ceiling, the receiver advertises `hevc` in `hello.codecs`, and the sender's own encoder supports it. HEVC pays extra encode cost, so it is used only where it buys something.
- The trigger condition rides on the decode-ceiling negotiation that #258 just shipped (`hello.maxEncodeWide/High`, PROTOCOL.md 6.5). The ceiling becomes codec-aware: the Mac receiver advertises 4096x2304 under H.264 and can raise it to its measured HEVC ceiling once HEVC is negotiated.

What #249 needs before it can carry this:

1. Rebase onto `Shared/StreamReceiver.swift` (it patches `iOS/PhoneReceiver.swift`, which moved in #145).
2. Change its selection logic from "HEVC whenever the receiver supports it" to the size-triggered rule above.
3. Verify the receiver drops every non-VCL HEVC NAL (access unit delimiter type 35, SEI 39/40) instead of feeding them to the decoder as slice data. #253 measured that getting this wrong corrupts every frame.
[...]

PR #249, opened 2026-08-27
feat(video): hardware-accelerated HEVC (H.265) encoding and decoding (#10)

kdbhalala:
## Summary

This PR implements hardware-accelerated HEVC (H.265) video encoding and decoding for higher visual quality at lower bitrates, especially beneficial over WiFi:

### 1. Hardware HEVC VideoToolbox Encoder on macOS (#10)
- Configures `kCMVideoCodecType_HEVC` compression sessions in `MacSender` with low-latency rate control and automatic fallback to H.264 if HEVC hardware compression is unavailable.
- Extracts and transmits HEVC parameter sets (VPS, SPS, PPS) in `annexB(from:)`.

### 2. HEVC Playback & Format Description on iOS (#10)
- Detects HEVC NAL unit types (VPS: 32, SPS: 33, PPS: 34) and constructs `CMVideoFormatDescriptionCreateFromHEVCParameterSets`.
- Feeds hardware-decoded HEVC sample buffers directly into `AVSampleBufferDisplayLayer`.

### 3. Codec Handshake & Negotiation
- Announces supported codecs (`["h264", "hevc"]`) in receiver hello handshake.

## Verification
- **Unit Tests**: Added `MacTests/HEVCEncodingTests.swift` verifying VideoToolbox HEVC compression session initialization, real-time flags, and fallback.
- **Builds & Tests**:
  - `xcodegen generate && xcodebuild test -project OpenSidecar.xcodeproj -scheme OpenSidecarMac -destination 'platform=macOS'` — 36/36 tests pass.
  - `xcodebuild build -project OpenSidecar.xcodeproj -scheme OpenSidecariOS -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO` — builds cleanly.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -79,6 +79,7 @@ struct PhoneInfo: Decodable {
                           // (PROTOCOL.md 6.4); probed for a cable upgrade
     let maxEncodeWide: Int?  // receiver's decode ceiling in pixels (PROTOCOL.md
     let maxEncodeHigh: Int?  //  6.5): cap the stream, keep the desktop size
+    let codecs: [String]? // supported video codecs, e.g. ["h264", "hevc"] (issue #10)
 
     var kind: String { device ?? "device" }
     var protocolVersion: Int { pv ?? WireProtocol.assumedWhenAbsent }
@@ -1850,14 +1851,14 @@ final class MacSender: NSObject, SCStreamOutput, SCStreamDelegate {
 
     /// Create the compression session into `encoder`, optionally requiring an
     /// encoder that supports low-latency rate control.
-    private func createCompressionSession(width: Int, height: Int, lowLatency: Bool) -> OSStatus {
+    private func createCompressionSession(width: Int, height: Int, codecType: CMVideoCodecType, lowLatency: Bool) -> OSStatus {
         let spec: CFDictionary? = lowLatency
             ? [kVTVideoEncoderSpecification_EnableLowLatencyRateControl: kCFBooleanTrue] as CFDictionary
             : nil
         return VTCompressionSessionCreate(
             allocator: nil,
             width: Int32(width), height: Int32(height),
-            codecType: kCMVideoCodecType_H264,
+            codecType: codecType,
             encoderSpecification: spec,
             imageBufferAttributes: nil,
             compressedDataAllocator: nil,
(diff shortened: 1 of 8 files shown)


peetzweg:
Direction for this PR is now settled, details in https://github.com/peetzweg/opendisplay/issues/10#issuecomment-5499339750. Short version: HEVC is wanted (the Mac receiver made 5K panels a real use case), but before merge this needs (1) a rebase onto Shared/StreamReceiver.swift, which replaced iOS/PhoneReceiver.swift in #145, (2) selection changed from 'HEVC whenever the receiver supports it' to size-triggered: H.264 stays default, HEVC only when the requested stream exceeds the H.264 decode ceiling from #258, (3) a check that every non-VCL HEVC NAL (AUD type 35, SEI 39/40) is dropped rather than fed to the decoder as slice data, per the field evidence in #253, and (4) end-to-end validation on real hardware.

peetzweg:
Brought this branch up to date and onto the policy from #10. The encode and packetize groundwork held up well and is still here, rebased; thank you for it. Three commits now sit on top of yours:

**Rebased onto main.** The receiver moved from `iOS/PhoneReceiver.swift` to `Shared/StreamReceiver.swift` in #145, so your decode path now serves the iPhone, the iPad and the Mac receiver app alike.

**Codec selection is size-triggered** (PROTOCOL.md 6.6, decided in #10). H.264 stays the default; the sender picks HEVC only when the requested stream exceeds the receiver's H.264 decode ceiling, the receiver advertises `hevc` in `hello.codecs`, and the sender's encoder accepts that size (probed with a throwaway session before the capture is configured). The Mac receiver advertises `codecs` plus a 5120x2880 HEVC ceiling where it has hardware HEVC decode; phones advertise nothing and stay on H.264 end to end. Your `-hevc YES` flag still forces the HEVC path for testing, now gated on receiver support. The negotiation lives in `Mac/StreamNegotiation.swift` with unit tests, and is applied on all three capture paths (extend, mirror, rotation).

**Fixed three decode bugs in the parsing path:**

1. Codec detection keyed on the shifted HEVC NAL type, but an H.264 non-IDR slice (0x41, every P-frame VideoToolbox emits) shares those bits with an HEVC VPS, so a plain H.264 session flipped into HEVC mode on its first delta frame and ate it as a VPS. Detection now uses bytes that are unambiguous in both codecs (0x40 for VPS, 0x67/0x68 for H.264 SPS/PPS).
2. Only SEI was skipped, but the hardware HEVC encoder emits an access unit delimiter (type 35) per frame, and #253 measured that feeding one to the decoder as slice data corrupts every frame. The parser now whitelists VCL and parameter sets and drops every other non-VCL unit, on the H.264 side too.
3. `isHEVC` latched on forever and survived disconnects; it now resets with the stream state and switches cleanly in both directions mid-connection.

Verified end to end on a Mac-to-Mac loopback session with `-hevc YES`: sustained 59 fps, 15 ms end to end (16 ms p95), decode time ~0 ms, and zero decode errors or keyframe re-requests over the whole run. The receiver builds the HEVC format description from VPS/SPS/PPS and decodes the real bitstream (per-frame access unit delimiter included) with no corruption.

Still open before this merges: validation at a real 5K panel, where the size trigger fires on its own rather than via the force flag. The 2017 iMac from #253 is the ideal machine for that.

PR #251, opened 2026-08-28
Add on-screen keyboard & fix sidebar modifier handling

Portgas443:
- **feat(input): hardware keyboard passthrough & on-screen modifier sidebar (#6, #7)**
- **Fix on-screen keyboard modifier handling**
## What changed
- Added on-screen keyboard handling
- Add Caps Lock toggle handling
- Preserve Caps Lock state when sidebar modifiers are updated
- Fix modifier handling between the iOS keyboard and Mac input injection
- Fix sidebar modifier handling

## Testing

- Verified the changes build and remain clean in the working tree
- Verified the changes are isolated to the keyboard/modifier functionality

### DIFF
--- a/Mac/InputInjector.swift
+++ b/Mac/InputInjector.swift
@@ -41,6 +41,7 @@ final class InputInjector {
     private let vendorPointerType: Int64 = 0x0802    // Grip Pen (what apps expect)
     private let capabilityMask: Int64 = 0x05C7       // pressure + tilt + rotation + buttons
     private var inRange = false
+    private var stickyModifiers: CGEventFlags = []
 
     // Pencil-only synthetic click counting — tablet events don't get click
     // state from the Window Server, so we mirror macOS double-click prefs here.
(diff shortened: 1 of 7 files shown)


peetzweg:
Interesting, what is the usecase here? I assume using the an iPad as a temporary monitor for a mac mini which does not even have a keyboard attached? 

That's actually a cool idea and probably super helpful. Will look into this once I can.

Portgas443:
> Interesting, what is the usecase here? I assume using the an iPad as a temporary monitor for a mac mini which does not even have a keyboard attached?
> 
> That's actually a cool idea and probably super helpful. Will look into this once I can.

I have an old trashcan mac that I log in via screen share bc it has no attached monitor. The idea was to use the iPad as a monitor from another room. But I realized it needs a monitor connected to work. It just shows a black screen when no monitor is connected. Maybe my old trashcan is the issue, I don't have a newer mac mini to test on. I may go back and see what I can do.

Works great on a mac pro, though, if you are in another room and need to type via ipad it works well.

Issue #241, opened 2026-08-26
Mac receiver: pick #145 over #198, ship as a separate low-floor target, evaluate browser receiver


peetzweg:
Two open PRs implement the Mac receiver (#82 / #17): https://github.com/peetzweg/opendisplay/pull/145 (in-app role switch, shared receiver core) and https://github.com/peetzweg/opendisplay/pull/198 (separate `OpenSidecarMacReceiver` target, forked receiver core, big sender changes). This issue records the comparison, the decision on which to carry forward, and two constraints we want the production version to meet: the receiver should run on as much legacy Mac hardware as possible, and we want to know whether a browser page can be a receiver too.

## Decision: continue #145's code in #198's shape

**Code from #145.** It moves `iOS/PhoneReceiver.swift` to `Shared/StreamReceiver.swift` (UIKit removed, device kind / fallback name / cursor sprite injected) and reuses it unchanged for the Mac. No sender changes, no wire changes (`pv` stays as is, `hello.device` says "Mac"). 6 files, +954/-332. The iOS receiver and the Mac receiver stay one implementation, which is the property we care most about for maintenance.

**Shape from #198.** A separate app target with its own bundle id and its own deployment target. Reason below (legacy floor). #145's runtime role switch would force the whole sender app to build at the lower floor, with `@available(macOS 14, *)` around every ScreenCaptureKit / `MenuBarExtra` / `Duration` / CGVirtualDisplay call site, plus split Sparkle appcasts. That is a permanent tax on every sender change.

**Why not merge #198 wholesale.** Reviewed on top of current `main`:
- `MacReceiver/DisplayReceiver.swift` (1336 lines) is `iOS/PhoneReceiver.swift` copied and extended; identical structure, function names and comments. Nothing moved to `Shared/`, so every receiver fix would land twice.
- `Mac/MacSender.swift` grows 1438 to 2489 lines and rewrites exactly the regions `main` changed since its merge-base (`setupEncoder` / `createCompressionSession` from #202, `captureGeneration` from #207, `VirtualDisplay.resize` from #210, poisoned-identity handling from #231). Rebase is a re-implementation.
[...]

PR #253, opened 2026-08-29
Add a macOS receiver: true 5K on an iMac 27" 2017, plus mirror and HEVC fixes

Venelinhr:
> Supersedes #252 — auto-closed when I renamed the source branch. Same content.

## Tested and validated

| | |
|---|---|
| **Resolution** | **5120 × 2880 — 14.7 MP, 1:1 pixels** |
| **Desktop** | 2560 × 1440 @2x |
| **Codec** | HEVC Main, 30 Mbit/s |
| **Latency** | 54 ms median, 62 ms p95 |
| **Round trip** | 5–6 ms |
| **Machine** | i5-7600, Radeon Pro 575, **Ventura 13.7.8**, **Wi-Fi** |

Sender was a MacBook Pro M4 Pro on macOS 26. Every codec and resolution limit
quoted below was established by **end-to-end playback on this hardware**, not from
spec sheets. Full setup and usage notes: [`MacReceiver/README.md`](MacReceiver/README.md).

---

## What this adds

A **macOS receiver**, so a Mac can be a second display for another Mac — not just an iPad or iPhone.

The machine this was built for is an **iMac 27" 2017 (Retina 5K)**. It has no other way to do this: Target Display Mode was removed after 2014, Apple capped it at macOS 13 Ventura (so it can't run the sender), and AirPlay-to-Mac needs a 2019+ iMac.

It now runs a **5120×2880 HEVC** stream on its 5120×2880 panel — 1:1 pixels, no scaling, ~50 ms end to end over Wi-Fi.

## Four bugs fixed

Three of these affect existing users, not just the new target.

**1. Mirror captured at half resolution** — `MacSender.swift`
`SCDisplay.width/height` report **points**, not pixels. On a Retina Mac that meant sending 1728×1116 from a screen macOS actually renders at 3456×2232. Half the detail was discarded before the frame ever left the sender, and no quality setting could recover it. Now reads `CGDisplayMode.pixelWidth/Height`.

**2. Desktop size was welded to stream size**
`points = announced pixels / 2`, and `encode = points × 2 × quality.scale`, so asking for a large desktop forced a proportionally large stream. A receiver that wants a big desktop but has a modest decoder had no way to say so. Added an additive **`hello.maxEncodeWide` / `maxEncodeHigh`** — PROTOCOL.md §6 makes unknown fields ignorable, so no `pv` bump is needed and older peers are unaffected.

**3. HEVC was emitted with no parameter sets at all**
`annexB()` looped `0..<2` using `CMVideoFormatDescriptionGetH264ParameterSetAtIndex`. HEVC carries **three** sets (VPS/SPS/PPS) and needs its own accessor — the H.264 one returns an error for an HEVC format description. The stream went out with no headers, so no decoder could ever start. This is what made 5K look like a hardware limit when it wasn't.

**4. Receiver treated every non-parameter NAL as picture data**
HEVC emits an access-unit delimiter (type 35) that H.264 does not. Feeding it to the decoder as slice data corrupts every frame. Only types 0–31 are VCL.

## New files

- **`MacReceiver/`** — the receiver app. AppKit, universal, deployment target 13.0. Reuses `iOS/PhoneReceiver.swift` and `iOS/MetalVideoRenderer.swift` **unchanged and in place**; the only new abstraction is `Receiver/Platform.swift`, the single file that branches on toolkit.
- **`MacSenderShim/`** — builds the sender with `swiftc` and no Xcode. Developed on a managed Mac with no admin rights, where the Xcode licence can never be accepted and `xcodebuild`/`xcodegen` are therefore unusable. `DEVELOPER_DIR` points at the Command Line Tools; Sparkle is stubbed behind `#if !canImport(Sparkle)`.
- **`tools/decode-probe.swift`** — reports which codec/size pairs a machine can actually hardware-decode, using **real captured parameter sets**. Worth noting: a format description built via `CMVideoFormatDescriptionCreate` with no parameter sets fails session creation on *every* Mac, so a probe written that way measures nothing. Mine did, at first.

## Measured, not assumed

- **H.264 hardware decode stops below 5120 wide on every Mac tested, including an M4 Pro.** A format ceiling, not an age one — 4K is H.264's practical limit here.
- **HEVC handles the full 5120×2880** on a 2017 iMac (Kaby Lake + Radeon Pro 575).
- `VTDecompressionSessionCreate` succeeding proves nothing on its own — it builds sessions the hardware cannot sustain. Only end-to-end playback is a valid test.

## Render path

Manually created `CALayer`s default to `contentsScale = 1.0` and do not inherit the window's backing factor the way iOS layers do. Left at 1 on a Retina panel the video is rasterised at point resolution and then stretched, which presents as *"sharp in a small window, blurry at full screen"*. Now set from `viewDidMoveToWindow` as well as `viewDidChangeBackingProperties`, with a screen-derived fallback, and logged.

The Metal renderer is also wired up on macOS with a tunable unsharp mask, for the case where the stream is smaller than the panel. At true 5K it is best left at `0`.

## Deliberately out of scope

**No input forwarding.** macOS Universal Control already shares keyboard and mouse between two Macs over Apple's own encrypted channel. Given PROTOCOL.md §1 ("no TLS and no authentication at `pv` 3"), keeping keystrokes off this wire seemed better than adding them.

## Notes for review

- `project.yml` is untouched — `MacReceiver/build.sh` lists sources directly, since `xcodegen` was not usable here. Adding a proper target would be a reasonable follow-up for anyone who has Xcode.
- The iOS app compiles against the shim changes but **could not be built or run** on this machine, for the same reason. The diff to `iOS/PhoneReceiver.swift` is small and mechanical (`UIImage` → `CGImage` in one closure signature, `UIDevice` → `ReceiverPlatform`), but it wants a real iOS build before merging.
- Sender settings used here: `hevc=YES`, `bitrateMbps=30`, `quality=best`, `pixfmt=bgra`. **30 Mbit/s is a ceiling** on this hardware — at 45 the receiver cannot sustain 5K decode over Wi-Fi and the watchdog walks the announced size down, which looks like "the UI got big" rather than an error.

### DIFF
--- a/Mac/MacSender.swift
+++ b/Mac/MacSender.swift
@@ -39,10 +39,18 @@ enum StreamQuality: String, CaseIterable {
     }
 
     var bitrate: Int {
+        // Tunable without a rebuild:
+        //   defaults write com.peetzweg.opensidecar.mac bitrateMbps -int 40
+        if let mbps = UserDefaults.standard.object(forKey: "bitrateMbps") as? Int, mbps > 0 {
+            return mbps * 1_000_000
+        }
+        // Raised from 18/10/6: mirror now captures at native pixel resolution,
+        // which is 4x the pixels of the old point-sized capture. At the old
+        // bitrate that traded softness for blockiness instead of fixing it.
         switch self {
-        case .best: return 18_000_000
-        case .balanced: return 10_000_000
-        case .fast: return 6_000_000
+        case .best: return 32_000_000
+        case .balanced: return 18_000_000
+        case .fast: return 9_000_000
         }
     }
 
@@ -73,9 +81,26 @@ struct PhoneInfo: Decodable {
                           // across USB and WiFi
     let pv: Int?          // receiver protocol version (issue #132); absent on
                           // every pre-handshake install → treat as protocol 1
+    // The widest/tallest frame this receiver can actually decode. Lets a
+    // receiver ask for a large DESKTOP without also demanding a stream its
+    // hardware chokes on — an iMac 27" 2017 wants 2560x1440 points but tops out
+    // near 3840x2160 of H.264. Optional, so older receivers are unaffected
+    // (PROTOCOL.md §6: unknown fields are ignored, so this needs no pv bump).
+    let maxEncodeWide: Int?
+    let maxEncodeHigh: Int?
 
     var kind: String { device ?? "device" }
     var protocolVersion: Int { pv ?? WireProtocol.assumedWhenAbsent }
+}
+
+/// Shrink a capture size to the receiver's declared decode ceiling, preserving
+/// aspect ratio. Without a ceiling this is just the existing even-rounding.
+func clampCapture(_ w: Int, _ h: Int, to info: PhoneInfo) -> (w: Int, h: Int) {
+    guard let maxW = info.maxEncodeWide, maxW > 0, w > maxW else {
+        return (w & ~1, h & ~1)
+    }
+    let scaledH = Int(Double(h) * Double(maxW) / Double(w))
+    return (maxW & ~1, scaledH & ~1)
 }
 
 /// How the sender reaches the receiver. Reconnects re-dial from scratch, so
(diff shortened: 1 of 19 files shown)


peetzweg:
Thanks for the contribution. I will look into this. I'm already working on a receiver app and will try to incorporate what you did here for the 5k iMac! 👍 

https://github.com/peetzweg/opendisplay/pull/145

Venelinhr:
Hi, thank you for this response.
You helped me to find solution to use my old iMac 27” as side screen,
finally 🙌

Thank you

peetzweg:
Thank you for this. The measurement discipline here is rare: real hardware, end to end playback as the only accepted evidence, and a probe tool that admits its own first version measured nothing. We learned things from this PR that we are acting on.

You had bad timing though, sorry. A standalone Mac receiver shipped in v1.18.0 two days after you opened this, built on #145 (the decision and constraints are in #241, including why we keep one receiver core shared between iOS and macOS instead of carrying a second implementation). So the receiver app half of this PR will not be merged, and I will close this one.

The rest is too good to lose, so it is landing:

- #258 lifts your mirror fix (points vs pixels) and hello.maxEncodeWide/High onto the shipped core, with credit. The Mac receiver now advertises 4096x2304 based on your H.264 ceiling measurements, so a 5K panel gets its full desktop over a 4K stream instead of a session the decoder cannot sustain.
- HEVC is in flight in #249 and already uses the HEVC parameter set accessor; your access unit delimiter finding (type 35 fed to the decoder as slice data) is now a review checkpoint there. Once that lands, your iMac should do true 1:1 5K.
- decode-probe.swift would be great as its own small PR into tools/, if you are up for it. That keeps your authorship on it, and the "session creation proves nothing, only sustained playback does" methodology is worth having in the repo.

Meanwhile: OpenDisplay Receiver 1.18.0 runs on macOS 12 and later, so your Ventura iMac is covered today. https://github.com/peetzweg/opendisplay/releases/tag/v1.18.0 has both apps. At 5K you will get a 4K stream upscaled until HEVC lands, but I would love to hear how it behaves on your hardware, and your iMac is exactly the validation machine #249 needs.

PR #275, opened 2026-09-09
fix(mac): pace large H.264 streams below the level limit

### LINKED ISSUE #271: Mac receiver on a 5K panel: 'Best' quality silently exceeds H.264's pixel-rate ceiling (encoder rejects every frame); 'Balanced' fixes it
## Summary

Mac→Mac receiver over Ethernet to a **27" 5K panel** (2015 iMac, Monterey 12.7.6 receiver) never shows video — receiver reports Connected/Receiving, but the desktop never renders, and the sender kills the session after ~1-2 minutes ("cable link lost (silent for >5s)"). Root cause is the encoder silently rejecting almost every frame at the "Best" quality preset, because the resulting pixel×fps rate exceeds H.264's Level 5.2 ceiling. **Switching the sender's Quality picker from "Best" to "Balanced" fixed it completely** — filing this so the next person with a 5K/large-panel Mac receiver doesn't have to rediscover it.

## Environment

| | |
|---|---|
| Sender | Mac mini, Apple M2 |
| Receiver | Late-2015 27" Retina 5K iMac, macOS Monterey 12.7.6 |
| Transport | Wired Ethernet, both machines on the same TP-Link Deco mesh node |
| OpenDisplay | Mac Receiver + Mac Sender, built on top of #145 (merged 2026-09-01) |

## Symptom

- Receiver panel shows **Connected**, then **Receiving 4096×2304** — but the video window never shows the extended desktop, and the receiving Mac's cursor doesn't move onto the display.
- After roughly 50-110 seconds, the session drops back to "waiting for a Mac" on the receiver, and the sender ends the session.
- This repeats on every reconnect attempt, identically.

## Root cause (found in the sender's own log, `~/Library/Logs/OpenDisplay/opendisplay.log`)

At "Best" quality, the encoder is asked to hit 4096×2304 (the receiver's `maxEncodeWide`/`maxEncodeHigh`, [`MacReceiver.swift:51`](https://github.com/peetzweg/opendisplay/blob/main/MacReceiver/MacReceiver.swift#L48-L51)) at the sender's hardcoded 60fps target (`kVTCompressionPropertyKey_ExpectedFrameRate = 60`). Every session shows this immediately after `capture started`:

```
[09:49:47.685] capture started: 4096x2304 display 11 generation 1 mode extend localCursor=true
[09:49:47.685] status[wifi:iMac 27]: Extending to Mac (4096×2304)
[09:49:47.755] encoder output rejected: nil buffer despite noErr (1 since last report)
[09:49:48.757] encoder output rejected: nil buffer despite noErr (54 since last report)
[09:49:49.757] encoder output rejected: nil buffer despite noErr (57 since last report)
```

This is exactly the failure mode described in #118 — `VTCompressionSessionEncodeFrame` returns `noErr` but hands back a nil buffer once the pixel-rate exceeds H.264 Level 5.2's ~535M luma samples/sec ceiling. The math:

**4096 × 2304 × 60fps = 566,231,040 samples/sec — about 6% over the ~535M ceiling.**

With almost no frame ever actually encoded, the receiver has essentially nothing real to decode (even though it can still parse one early SPS/PPS pair and report a "Receiving WxH" status), and eventually the sender's 5-second silence watchdog fires and tears the whole session down:

```
[10:00:00.054] encoder output rejected: nil buffer despite noErr (1 since last report)
[10:00:00.444] cable link lost (silent for >5s) — unplugging means disconnect, ending session
[10:00:00.444] device disconnected — session wifi:iMac 27 stopped
```

Four separate reconnect attempts in the log all reproduce identically.

## What fixed it

Switching the sender's **Quality** picker from **Best (native)** to **Balanced (75%)** immediately fixed it — full stop, working extended desktop, no further drops. At 75% capture scale the effective resolution comes in well under the Level 5.2 ceiling even at 60fps, so the encoder stops rejecting frames.

## Suggested fix

"Best" quality is a trap for any receiver panel large enough to push width×height×60fps over ~535M — a real, silent failure with no user-facing signal beyond "nothing renders." A few options that would have caught this automatically:
- Clamp the default quality preset (or the effective capture resolution) so it can never exceed the H.264 Level 5.2 ceiling at the sender's target frame rate, regardless of which preset the user has selected.
- Surface the `encoder output rejected` counter somewhere user-visible (the performance overlay already exists and seems like a natural place) so "why is nothing rendering" is diagnosable without digging into `~/Library/Logs`.
- At minimum, mention this interaction in `COMPATIBILITY.md` or the receiver's own UI for large/5K panels, since 4096×2304 was chosen for *decode* reasons ([comment in `MacReceiver.swift`](https://github.com/peetzweg/opendisplay/blob/main/MacReceiver/MacReceiver.swift#L43-L47)) without accounting for the *encode*-side pixel-rate ceiling at 60fps.

Happy to test a patch if useful — the workaround (Balanced) is good enough for daily use in the meantime.

vipstanley:
Best quality can initialize a 4096×2304 Mac-receiver stream at 60 fps. On the reported 2015 5K iMac setup, VideoToolbox then returns `noErr` with a nil output buffer for nearly every frame, leaving the receiver effectively frozen. This change keeps the 4096×2304 raster and selects a conservative 55 fps target from the H.264 Level 5.2 macroblock-rate budget.

What changes:

- compute a rate policy from the encoded raster's 16×16 macroblocks;
- use the selected rate for `ExpectedFrameRate`;
- gate every encoder entry path, including replay and reconnect submissions;
- reuse the existing latest-frame replay timer when pacing defers a frame, so a final static-screen update is eventually delivered;
- keep existing 60 fps behavior for smaller rasters that fit the budget.

Before:

```text
encoder ready: 4096x2304 H.264 18Mbps quality=best
encoder output rejected: nil buffer despite noErr
```

After:

```text
H.264 L5.2 rate cap: 4096x2304 -> 55fps
encoder ready: 4096x2304 H.264 18Mbps fps=55 quality=best
```

Validation:

- Xcode 26.6 builds successfully on an M4 MacBook Pro running macOS 26.3.
- All 44 macOS unit tests pass, including ten new rate-policy tests.
- The completed public code was independently tested by the issue reporter using an M2 Pro Mac mini on macOS 26.5.2 and a 2015 5K iMac on macOS Monterey 12.7.6 over wired Ethernet. Best quality at 4096×2304 remained connected for several minutes without a drop.
- That test logged two isolated nil-buffer events during encoder startup and none persisted afterward. Delivered frame rate varied from roughly 15–57 fps, with occasional brief stalls.
- For terminal and web-browsing use, the reporter found the image noticeably sharper than Balanced quality and did not perceive added latency.
- A separate M4-to-2015-iMac Wi-Fi run delivered roughly 35–40 fps and likewise avoided the sustained near-every-frame rejection pattern.

Limitations:

- This does not enable native 5120×2880 streaming; it targets the existing 4096×2304 receiver ceiling.
- It does not claim zero dropped frames or universal compatibility across every encoder, display, and transport.
- Balanced remains the fallback for untested hardware.

The English build, rollback, log-checking, and feedback guide is in `5K_RECEIVER_WORKAROUND.md`.

Fixes #271.

### DIFF
--- a/5K_RECEIVER_WORKAROUND.md
+++ b/5K_RECEIVER_WORKAROUND.md
@@ -1,4 +1,4 @@
-# Experimental workaround for Best quality on a 5K Mac receiver
+# Proposed fix for Best quality on a 5K Mac receiver
 
 This branch proposes a sender-side workaround for [issue #271](https://github.com/peetzweg/opendisplay/issues/271). It preserves a 4096×2304 stream and reduces the nominal encoder frame rate to 55 fps. It does **not** enable native 5120×2880 streaming or guarantee 55 delivered fps.
 
@@ -70,7 +70,9 @@ To revert, quit OpenDisplay Dev and open your normal OpenDisplay sender. Balance
 
 - The initial prototype was built on an M4 MacBook Pro running macOS 26.3 with Xcode 26.6 and used with a 2015 5K iMac over Wi-Fi. It established a 4096×2304 extended stream, sustained approximately 35–40 delivered fps in the observed intervals, and reconnected successfully.
 - The prototype still logged isolated `nil buffer despite noErr` events at startup/reconnect. The sustained near-every-frame rejection pattern did not recur during the short observed sessions. This is not a claim of zero dropped frames or a long-duration soak test.
-- This public revision additionally routes replay/reconnect frames through the limiter and schedules delivery of a deferred final update. It has been built and has passed all 44 macOS unit tests, including ten new rate-policy tests. The completed public revision has **not yet had a fresh end-to-end receiver test**; the above hardware measurements belong to the initial prototype.
+- The completed public revision was independently tested by the issue reporter using an M2 Pro Mac mini on macOS 26.5.2 and a 2015 5K iMac on macOS Monterey 12.7.6 over wired Ethernet. Best quality at 4096×2304 remained connected for several minutes without a drop. Two isolated nil-buffer events occurred during encoder startup and none persisted afterward. Delivered frame rate varied from roughly 15–57 fps, with occasional brief stalls.
+- In that everyday-productivity test, the reporter found the image noticeably sharper than Balanced quality and did not perceive added latency. The workload covered terminal windows and web browsing rather than video playback or another stress test.
+- The public revision routes replay/reconnect frames through the limiter and schedules delivery of a deferred final update. It has been built and has passed all 44 macOS unit tests, including ten new rate-policy tests.
 - Unit tests cover rate selection, macroblock rounding, a 60 Hz source paced to 55 fps, deferred-frame eligibility, long idle gaps, reset, invalid timestamps, early admission, and duplicate prevention near a later slot. They do not emulate VideoToolbox or prove actual timer delivery to a physical receiver.
 - The workaround does not change receiver decode limits, codec, bitrate, or network handling. Lowering a nominal frame rate does not guarantee compatibility with every encoder, display, or transport.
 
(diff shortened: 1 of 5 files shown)


peetzweg:
Thanks for the focused implementation and the real-hardware validation on the 2015 5K iMac. This is now superseded by merged PR #288: the merged configuration path selects a capability-safe H.264 operating point, including 4096×2304 at 55 FPS for this case, and applies pacing across initial capture, reconfiguration, reconnect, and replay. Issue #271 was closed by #288, so I’m closing this PR as superseded.

PR #276, opened 2026-09-10
feat(mac): user-configurable frame rate (30/60/90/120 FPS)

KareemmSayed:
Adds a **Frame Rate** picker to the Mac app's settings with four presets:

| Preset | Use case |
|---|---|
| 30 FPS (Low Power) | Static content, saving battery |
| 60 FPS (Default) | Standard smoothness — the previous behavior |
| 90 FPS | High refresh, moderate overhead |
| 120 FPS (ProMotion) | Ultra-smooth on iPad Pro / iPhone Pro |

## Why

The capture pipeline was hardcoded to 60 FPS. Devices with 90/120 Hz panels
couldn't make use of their refresh rate, and users who prefer battery life
had no way to slow the stream down.

## How

- New `FrameRate` enum with a bitrate helper that auto-scales quality
  presets (30 FPS → 0.75×, 90 FPS → 1.25×, 120 FPS → 1.6×, e.g. Best at
  120 FPS ≈ 28.8 Mbps) so higher frame rates don't look blurry.
- The virtual display (`CGVirtualDisplay` mode and HiDPI recovery),
  ScreenCaptureKit pacing (`minimumFrameInterval`), and the
  `VTCompressionSession` (expected frame rate, average bitrate, keyframe
  interval) all read from the selected value.
- Selecting a new rate saves to `UserDefaults` and restarts running
  sessions, matching the existing "Show app in" / quality pickers.

Mac sender only — the phone receiver and the wire protocol are unchanged,
so this PR touches no shared protocol code.

## Phone-side tip: enable the Metal renderer at 90/120 FPS

At high frame rates the default video output (the system
`AVSampleBufferDisplayLayer`) can feel laggy: it paces presentation like a
media player and buffers frames internally, so a fast, bursty stream queues
up inside the layer and latency grows.

The phone app's Settings has a **Metal renderer (experimental)** toggle that
decodes with `VTDecompressionSession` and presents the *newest* frame
directly to a `CAMetalLayer` (latest-frame-wins, no queue), keeping latency
bounded at 90/120 FPS.

- **30/60 FPS:** leave it off — at 60 FPS the system video layer measured
  faster in this project's A/B tests (dedicated compositor plane).
- **90/120 FPS:** turn it on if the stream feels laggy.
- 90/120 FPS is best over USB — the 120 FPS "Best" preset runs ~28.8 Mbps,
  more than many WiFi links sustain.

Measured on-device with the Metal renderer at 120 FPS: median capture→glass
latency ~45–50 ms (`ph50`) even during delivery bursts.

## Testing

- New unit tests (`FrameRateTests`): raw values, labels, bitrate scaling
  (120 FPS Best = 28,800,000), and raw-value fallback.
- All 38 Mac tests pass.
- Manual: USB session with an iPhone Pro at each preset; the encoder log
  line now reports the rate, e.g.
  `encoder ready: 2556x1176 H.264 28Mbps @120fps quality=best`.

### DIFF
--- a/Mac/FrameRate.swift
+++ b/Mac/FrameRate.swift
@@ -1,0 +1,85 @@
+import Foundation
+
+/// Capture-resolution / bitrate trade-off. The virtual display always runs at
+/// native size — only the captured/encoded stream is scaled, so lower presets
+/// cut encode, transmit, and decode time at the cost of sharpness.
+enum StreamQuality: String, CaseIterable {
+    case best, balanced, fast
+
+    var scale: Double {
+        switch self {
+        case .best: return 1.0
+        case .balanced: return 0.75
+        case .fast: return 0.5
+        }
+    }
+
+    var bitrate: Int {
+        switch self {
+        case .best: return 18_000_000
+        case .balanced: return 10_000_000
+        case .fast: return 6_000_000
+        }
+    }
+
+    var label: String {
+        switch self {
+        case .best: return "Best (native)"
+        case .balanced: return "Balanced (75%)"
+        case .fast: return "Fast (50%)"
+        }
+    }
+
+    var explanation: String {
+        switch self {
+        case .best: return "Pixel-perfect at the device's native resolution. Highest bandwidth and latency."
+        case .balanced: return "75% capture resolution — noticeably lower latency, slight softness."
+        case .fast: return "Half resolution — lowest latency and bandwidth, visibly softer. Good for WiFi."
+        }
+    }
+}
+
+/// User-selectable target frame rate for the capture stream and virtual display.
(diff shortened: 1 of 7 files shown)


peetzweg:
PR #288 has landed the codec and receiver capability safety limits, but this PR’s user-selectable 30/60/90/120 FPS preference remains separate and useful. Please rebase or port it onto current `main`. The effective rate should be the minimum of the user’s preference, the receiver’s advertised display/decoder limit, and the codec-safe rate selected by `H264StreamConfiguration`; the preference must not bypass the 4096×2304 H.264 cap. The existing Best/Balanced/Fast choices remain the resolution/bitrate intent.

PR #281, opened 2026-09-13
Recover Mirror capture and add Pause/Resume

### LINKED ISSUE #280: Mirror mode does not recover after ScreenCaptureKit stops the stream
## Problem

In Mirror mode, the video stream can occasionally freeze while the receiver remains connected and input continues to work.

Disconnecting and reconnecting restores video.

The logs show:

```text
SCStreamErrorDomain Code=-3821
"Stream was stopped by the system"
```
## Current behavior

The existing capture-stop handler recovers unexpected stream termination in Extend mode, but not in Mirror mode.

The recovery path is currently gated by :

```swift
guard !stopped, mode == .extend else { return }
```

As a result:

Extend mode attempts to reattach or rebuild capture
Mirror mode returns without recovery
the receiver connection remains alive
input can continue working
the last video frame remains frozen until the session is disconnected and reconnected
## Expected behavior

Unexpected ScreenCaptureKit termination should use the existing bounded recovery flow in Mirror mode as well.

If the mirrored physical display is still available, capture should be reattached/restarted without requiring a full disconnect.

Intentional stops should remain distinct and must not trigger automatic recovery.

## Notes

This appears to be a capture-lifecycle issue rather than a transport disconnect, since the control/input connection remains active while video is frozen.

raiseCatError:
Fixes #280.

An unexpected ScreenCaptureKit termination could leave Mirror sessions with frozen video while their connection and input remained alive. Mirror now uses bounded recovery alongside Extend: it first reattaches capture to the existing display and rebuilds the capture pipeline when reattachment fails. Intentional capture stops remain separate and never trigger automatic recovery.

This also adds Pause/Resume controls for active Mirror and Extend sessions. Pause keeps the receiver connected, releases and temporarily gates input, and shows a lightweight “Display Paused” overlay. Resume restarts capture without reconnecting.

The sender synchronizes `displayState=running` after every successful capture start, including mode replacement and recovery, preventing the receiver from retaining stale paused UI.

Tested manually on a real iPhone in Mirror and Extend modes, including repeated Pause/Resume, disconnect while paused, input gating, and switching modes while paused. A real `-3821` event did not recur during final testing; recovery decisions, lifecycle transitions, retry bounds, and display-state synchronization are covered by focused tests.

Validation:
- 40 macOS tests passed
- macOS Debug build passed
- unsigned generic iOS build passed
- `git diff --check` passed


peetzweg:
PR #288 now contains bounded Extend recovery, transactional virtual-display resizing, and stricter cancellation/connection-generation handling. It does not add Mirror recovery or Pause/Resume, so issue #280 and this PR remain relevant. Please rebase or port this onto current `main`, preserving the Mirror-specific recovery and pause state while using the new shared capture/configuration lifecycle rather than restoring the older Extend path.

PR #290, opened 2026-09-17
test(tools): add a hardware decode capability probe

Venelinhr:
Splitting this out of #253 as you suggested — it stands on its own and does not
depend on anything else from that branch.

## What it is

`tools/decode-probe.swift` answers one question on the machine you run it on:
**which codec and size pairs can this hardware actually decode?**

```
MACHINE: Venelin's iMac
MODEL:   iMac18,3
CPU:     Intel(R) Core(TM) i5-7600 CPU @ 3.50GHz

Hardware decode support on this Mac
===================================
  H.264  3840x2160  (4K)   OK   [3840x2160]
  H.264  5120x2880  (5K)   NOT SUPPORTED   [5120x2880]
  HEVC   3840x2160  (4K)   OK   [3840x2160]
  HEVC   5120x2880  (5K)   OK   [5120x2880]
===================================
HEVC 5K decodes -> true 5K is worth implementing.
```

No build step, no Xcode: `swift tools/decode-probe.swift`.

## Why the parameter sets are checked in

This is the part worth keeping, and it is why the tool exists in this shape.

A `CMVideoFormatDescription` built with `CMVideoFormatDescriptionCreate` and **no
parameter sets** fails `VTDecompressionSessionCreate` on *every* Mac — including
ones that decode the format perfectly well. A probe written that way reports
"not supported" for everything and measures nothing.

My first version did exactly that, and I believed it for a while. It is the
reason I spent a day convinced the 2017 iMac could not do HEVC 5K, when the real
fault was in my own `annexB()`.

So the headers here are **real SPS/PPS — and VPS for HEVC — captured from a live
encoder** by the companion `tools/make-probe-headers.swift`. That also buys the
thing that matters on old hardware: the probe can run on a machine that cannot
**encode** the size it is being asked to **decode**.

## The caveat, stated in the source

`VTDecompressionSessionCreate` succeeding proves the decoder can be *built*, not
that the hardware can *sustain* it. At 45 Mbit/s my iMac built a 5K session and
then could not keep up, which presented as the UI getting bigger rather than as
an error. Only end-to-end playback settles it. The comment says so, so nobody
reads a green line here as a promise.

## Measured so far

| Machine | H.264 5K | HEVC 5K |
|---|---|---|
| iMac 27" 2017 — i5-7600, Radeon Pro 575 | no | **yes** |
| MacBook Pro M4 Pro | **no** | yes |

H.264 hardware decode failing at 5120 wide on an **M4 Pro** is the result I did
not expect. It is a format ceiling, not an age one — which is what #258 now
encodes as the receiver's 4096-wide advertisement.

## Notes

- Two files: the probe, and the generator that produced its headers. The
  generator only needs re-running if new sizes are added, but it documents where
  the base64 came from, which seemed better than magic constants.
- Sits beside `tools/fake-receiver.swift` and follows it — plain `swift file.swift`,
  no target, no project changes.
- Happy to drop `make-probe-headers.swift` if you would rather keep `tools/` lean.


Venelinhr:
Closing this one — withdrawing it for now.

